PR #345, opened 2020-11-30
chore: replace tslint with eslint

dbjorge:
#### Description of changes

This PR replaces all usage of tslint with eslint, per current SDL requirements and tslint's deprecation. I started from a clean eslint configuration using the SDL recommendations, rather than keeping the existing tslint rule suppressions, since this codebase is small enough to just fix everything during the conversion.

#### Pull request checklist


src/progress-reporter/check-run/check-run-creator.spec.ts
```diff
@@ -114,20 +112,19 @@ describe(CheckRunCreator, () => {
                 title: checkRunDetailsTitle,
                 summary: disclaimerText,
                 annotations: [],
-                text: stripIndent`
-                ${message}`,
+                text: stubErrorMarkdown,
```
peterdur, line 115: No longer need `stripIndent` import above after this change

peterdur: (and then can remove packages `common-tags` and `@types/common-tags`

dbjorge: Leaving no-unused-vars enabled in the tests would have actually caught this, hah! I think I'll re-enable it and do inline suppressions

peterdur (approved):
Looks good!

PR #399, opened 2021-01-06
fix: prevent error when injecting axe.min.js introduced in #395

dbjorge:
#### Description of changes

Fixes a bug introduced in #395 where resolving axe.min.js doesn't work correctly when the bundled index.js is run from a working directory that isn't its own directory (because `require.resolve` returns a relative path as #395 used it).

#### Pull request checklist


lisli1 (approved):
LGTM

PR #459, opened 2021-02-16
Using resolutions to fix marked Dependabot alert

AhmedAbdoOrtiga:
#### Details

Dependabot cannot update marked to a non-vulnerable version
The latest possible version that can be installed is 1.2.9 because of the following conflicting dependency:

accessibility-insights-scan@0.6.0 requires marked@^1.1.0 via a transitive dependency on apify-shared@0.5.0
The earliest fixed version is 2.0.0.

#### Pull request checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
         "serialize-javascript": ">=3.1.0",
         "apify-shared": ">=0.5.0",
         "y18n": ">=5.0.5",
-        "socket.io": ">=2.4.0"
+        "socket.io": ">=2.4.0",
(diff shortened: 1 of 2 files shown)


dbjorge (changes requested):
The package.json/yarn.lock changes look fine, but it's important that we don't update dist/index.js without also doing a corresponding update to NOTICES.txt. I'd recommend splitting out the index.js update to a separate "release" PR.

dbjorge (approved):

PR #558, opened 2021-04-28
feat(gh-action-crawling): add scanning of remote URL

lisli1:
#### Details

* Added `url` input to action.yml and corresponding getUrl to task-config.ts
* Modified scanner.ts to use remote url as the scan url if `url` is specified in the input. If `url` is not specified, start the file server and use scan url relative path for the scan url.

##### Motivation

Addresses WI: # 1837995

##### Context

Verified the following scenarios:
* [Added `url` input in test repo](https://github.com/lisli1/test-action/runs/2452665414?check_suite_focus=true) and checked that the resulting artifact scan report and an exported report from running FastPass on the same URL had the same results
* [Removed `url` from test repo](https://github.com/lisli1/test-action/runs/2452718082?check_suite_focus=true) to check that the local file is scanned when the `url` input is not present
* [Added empty string `''` for `url` input in test repo](https://github.com/lisli1/test-action/runs/2452730011?check_suite_focus=true) and verified that the local file is scanned instead of trying to scan a non-existing remote URL

#### Pull request checklist

- [x] Addresses an existing issue: # 1837995

### DIFF
--- a/src/scanner/scanner.ts
+++ b/src/scanner/scanner.ts
@@ -40,8 +40,15 @@ export class Scanner {
 
         try {
             await this.allProgressReporter.start();
-            const baseUrl = await this.fileServer.start();
-            scanUrl = url.resolve(baseUrl, this.taskConfig.getScanUrlRelativePath());
+
+            const remoteUrl: string = this.taskConfig.getUrl();
+            if(remoteUrl) {
+                scanUrl = remoteUrl;
+            }
+            else {
+                const baseUrl = await this.fileServer.start();
+                scanUrl = url.resolve(baseUrl, this.taskConfig.getScanUrlRelativePath());
+            }
 
             this.logger.logInfo(`Starting accessibility scanning of URL ${scanUrl}.`);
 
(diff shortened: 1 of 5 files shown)


src/scanner/scanner.ts
```diff
@@ -40,8 +40,14 @@ export class Scanner {
...
         try {
             await this.allProgressReporter.start();
-            const baseUrl = await this.fileServer.start();
-            scanUrl = url.resolve(baseUrl, this.taskConfig.getScanUrlRelativePath());
+
+            const remoteUrl: string = this.taskConfig.getUrl();
+            if (remoteUrl) {
+                scanUrl = remoteUrl;
+            } else {
+                const baseUrl = await this.fileServer.start();
```
karanbirsingh, line 48: I think we have a `fileServer.close()` later on in the code. Maybe we can change the implementation of `close` so that it's a no-op if we didn't call `start` - or making sure it doesn't get called if we don't call start, etc.

karanbirsingh: @lisli1 pointed out we do this already with `startServerPromise`, but I missed it! Resolving this comment

karanbirsingh (approved):

PR #564, opened 2021-04-29
feat(gh-action-crawling): add crawling options to action input metadata

karanbirsingh:
#### Details

This PR adds crawling options similar to [accessibility-insights-scan](https://www.npmjs.com/package/accessibility-insights-scan). It adds:
- url
- maxUrls
- discoveryPatterns
- inputFile
- inputUrls

tested [here](https://github.com/karanbirsingh/test-a11y-action/runs/2463027566?check_suite_focus=true) (action accepts `max-urls` in test)

##### Motivation

We want to pass these along to the crawler.

##### Context

In the CLI we do some parsing and error checking:
- https://github.com/microsoft/accessibility-insights-service/blob/main/packages/cli/src/cli.ts#L28
- https://github.com/microsoft/accessibility-insights-service/blob/main/packages/cli/src/crawler-parameters-builder.ts

We might want to expose some of this from AICrawler and reuse it. Otherwise in a follow-up PR we can reimplement some of these error checks.

#### Pull request checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -29,7 +29,7 @@ inputs:
         required: false
         default: '100'
     discovery-patterns:
-        description: 'List of URLs to crawl in addition to URLs discovered from crawling the provided URL, separated by space.'
+        description: 'List of RegEx patterns to crawl in addition to the provided URL, separated by space.'
         required: false
     input-file:
         description: 'File path that contains list of URLs (each separated by a new line) to scan in addition to URLs discovered from crawling the provided URL.'
--- a/action.yml
+++ b/action.yml
@@ -22,7 +22,20 @@ inputs:
         description: 'github token'
         required: true
     url:
-        description: 'remote site url'
+        description: 'The hosted URL to scan/crawl for accessibility issues.'
+        required: false
+    max-urls:
+        description: 'Maximum number of pages opened by crawler. The crawl will stop when this limit is reached.'
+        required: false
+        default: '100'
+    discovery-patterns:
+        description: 'List of URLs to crawl in addition to URLs discovered from crawling the provided URL, separated by space.'
+        required: false
+    input-file:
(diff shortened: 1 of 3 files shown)


action.yml
```diff
@@ -22,7 +22,20 @@ inputs:
...
     url:
-        description: 'remote site url'
+        description: 'The hosted URL to scan/crawl for accessibility issues.'
+        required: false
+    max-urls:
+        description: 'Maximum number of pages opened by crawler. The crawl will stop when this limit is reached.'
+        required: false
+        default: '100'
+    discovery-patterns:
+        description: 'List of URLs to crawl in addition to URLs discovered from crawling the provided URL, separated by space.'
```
lisli1, line 32: Is this a typo? Based on https://github.com/microsoft/accessibility-insights-service/blob/main/packages/cli/src/cli.ts#L102, seems like this should read 'List of RegEx patterns to crawl in addition to the provided URL, separated by space.'

src/task-config.spec.ts
```diff
@@ -4,90 +4,46 @@ import 'reflect-metadata';
...
+        'parses "$inputValue" as $expectedValue when provided as $inputOption',
+        ({ inputOption, getInputFunc, inputValue, expectedValue }) => {
+            actionCoreMock
+                .setup((am) => am.getInput(inputOption))
+                .returns(() => inputValue)
+                .verifiable(Times.once());
+            const retrievedOption = getInputFunc();
+            expect(retrievedOption).toStrictEqual(expectedValue);
+        },
+    );
```
lisli1, line 46: I like that this removes a lot of redundancy for testing each getter!

action.yml
karanbirsingh: fixed, good catch!

lisli1 (approved):
LGTM

PR #566, opened 2021-04-29
Crawler integration

lamaks:
#### Details

- Convert action to use a composite run steps. The action will install npm dependencies on a runner when executing. The hard coded dependencies are not necessary
- Switch to use CLI crawler API
- Added test website to support local environment run
- Added support to run under local act environment

#### Pull request checklist


src/scanner/scanner.spec.ts
```diff
@@ -1,178 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
lamaks, line 1: Will add back in next PR

webpack.config.js
```diff
@@ -19,7 +17,14 @@ module.exports = (env) => {
         // We special case MPL-licensed deps because we want to avoid including their source in
```
karanbirsingh, line 17: It could be helpful for future reader to know this MPL comment only applies to `axe-core` and `@axe-core/puppeteer`, and other packages are external for different reason. Line 19 may need to change since PR removed copyWebpackPlugin

lamaks: Updated comment

karanbirsingh (approved):

PR #573, opened 2021-04-30
Added Windows development support instructions

lamaks:
#### Details

Added setup instruction to run action on Windows development environment

#### Pull request checklist

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -13,13 +13,9 @@ Licensed under the MIT License.
 
     ```
     dism.exe /online /enable-feature /featurename:Microsoft-Windows-Subsystem-Linux /all /norestart
-    ```
 
-    ```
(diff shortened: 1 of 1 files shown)


karanbirsingh (approved):
I was able to get this working with some slight variations:
- I also installed node & yarn on Ubuntu
- I switched to command prompt for now because my Windows terminal wasn't recognizing wsl ([docs](https://docs.microsoft.com/en-us/windows/wsl/install-win10#troubleshooting-installation))
- Windows users can maybe skip homebrew steps since they need to install from source anyways
- I pointed `act` to the local built version instead of homebrew

Then `yarn act:run` worked for me.

The act readme says Windows compat is [blocked](https://github.com/nektos/act#known-issues) in their readme, so this wsl approach is helpful! Once act releases their fix in 567 this process should also work outside wsl.

PR #585, opened 2021-05-05
feat(gh-action-crawling): hook up action inputs with AICrawler.crawl()

karanbirsingh:
#### Details

This PR hooks up the inputs in `action.yml` with `AICrawler.crawl()`. After this PR the action can be used with `input-urls`, `input-file`, etc.

validation:
- using input-urls: https://github.com/karanbirsingh/accessibility-insights-docs/runs/2512554417?check_suite_focus=true
- using input-file: https://github.com/karanbirsingh/accessibility-insights-docs/runs/2512515553?check_suite_focus=true

##### Motivation

Allow users to use crawling options

##### Context

Some options like `input-urls` and `discovery-patterns` will not work if the target url is locally hosted (e.g. `url` isn't defined`) because they assume absolute paths. This will be fixed separately.

#### Pull request checklist


src/scanner/crawl-argument-handler.ts
```diff
@@ -33,17 +33,21 @@ export class CrawlArgumentHandler {
     }
 
     private getInitialScanArguments(): ScanArguments {
+        const inputFile = this.taskConfig.getInputFile() || undefined;
```
karanbirsingh, line 36: GitHub action does not distinguish between 'not provided' and 'empty string' so we need to explicitly map certain values to `undefined`: https://github.com/actions/toolkit/issues/272

If we don't do this, `crawlerParametersBuilder.build()` fails because passing an empty string for `input-urls` etc to the CLI is invalid.

lamaks (approved):

PR #592, opened 2021-05-06
Added support of fixed localhost port

lamaks:
#### Details

Added support of specify preferred localhost TCP port to use by website service
Remove node --prefix option
Fixed npm alert

#### Pull request checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -46,6 +46,7 @@ runs:
     steps:
         - name: Install npm packages
           run: |
+              echo 'Installing npm packages...'
               cd ${{ github.action_path }}/dist && npm install
               echo 'npm packages successfully installed'
           shell: bash
--- a/action.yml
+++ b/action.yml
@@ -38,12 +38,15 @@ inputs:
     input-urls:
         description: 'List of URLs to crawl in addition to URLs discovered from crawling the provided URL, separated by space.'
         required: false
+    localhost-port:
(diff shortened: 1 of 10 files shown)


action.yml
```diff
@@ -38,12 +38,16 @@ inputs:
...
+        description: 'The preferred local website TCP port to use when scanning local website content.'
+        required: false
 runs:
     using: 'composite'
     steps:
         - name: Install npm packages
           run: |
-              npm --prefix ${{ github.action_path }}/dist install
+              echo 'Installing npm packages...'
+              cd ${{ github.action_path }}/dist && npm install
```
karanbirsingh, line 50: It fixed --prefix issue, but it failed later on for windows-2019. It works if I:
- replace all instances of `${{ github.action_path }}` with `$GITHUB_ACTION_PATH` like [here](https://github.community/t/github-action-path-contains-on-windows-runners-even-when-the-shell-is-set-to-bash/126472)
- for windows, `scan-url-relative-path: index.html` must be added. Default value leads to scanning "c:/Program%20Files/Git/"

Second one is easy to workaround for user

karanbirsingh (approved):
Approving because port works correctly & ubuntu behavior is still correct. Windows behavior is broken for a different reason, I left a comment. We can fix now or separate PR

Issue #674, opened 2021-06-17
Angle brackets in rule description not escaped in scan result markdown


peterdur:
Visit https://github.com/asksep/test-ai-action/runs/2657373953?check_suite_focus=true

Note the third failed instance on this page reads

1 × image-alt: Ensures <img> elements have alternate text or a role of none or presentation

It should read

1 × image-alt: Ensures \<img\> elements have alternate text or a role of none or presentation

Viewing the page source finds a linked image between Ensures and elements:

`1 × image-alt</strong>:  Ensures <a target="_blank" rel="noopener noreferrer" href=""><img style="max-width:100%;"></a> elements` 

Running AI-Web on this page, entertainingly enough, spots an **image-alt** violation (the generated image has no alternate text or role) and a **link-name** violation (images in markdown are links to the image file). The empty linked image also creates an invisible tab stop.

Markdown generation should be reviewed to identify any other data or characters that need to be escaped.

karanbirsingh:
This has been released in v2.2.0

peterdur:
Verified corrected behavior at https://github.com/peterdur/test-ai-action/pull/3

PR #729, opened 2021-07-21
fix: escape special characters in markdown

karanbirsingh:
This PR escapes the < character in axe-core rule IDs and descriptions. The GitHub action (and later, the ADO extension) both render markdown. Without this change, strings like `` or `<video>` are (unsuccessfully) rendered as actual HTML images or videos in both GitHub and Azure DevOps.

#### Details

We replace instances of `<` with `\<` so that tags are not rendered in their HTML forms. GitHub describes some backslash-escaped values [here](https://enterprise.github.com/downloads/en/markdown-cheatsheet.pdf). While HTML tags aren't listed, experimentally this change works well. Azure DevOps [also describes backslash-escaped values](https://docs.microsoft.com/en-us/azure/devops/project/wiki/markdown-guidance?view=azure-devops#ignore-or-escape-markdown-syntax-to-enter-specific-or-literal-characters).

This PR only escapes the `<` character to keep scope small. To help guard against other unexpected renderings (imagine axe-core descriptions with hashtags or asterisks), we [generate a markdown snapshot file](https://github.com/microsoft/accessibility-insights-action/blob/cd14d867721e61155bcf8ee369e537e3d890ecd7/src/mark-down/__custom-snapshots__/axe-descriptions.snap.md) with all possible axe-core IDs and descriptions. Each time this file changes, reviewers can check whether we require changes to markdown generation.

After this change, running automated checks on the generated markdown UI shows no errors.

##### Motivation

Addresses issue #674 

##### Validation

The comment on this PR should properly render ``. I've added a rule failure to the demo website so every PR should produce this error. This PR also adds a snapshot.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #0000
- [x] (after PR created) The `Accessibility Checks (pull_request)` check should fail. All other checks should pass.

### DIFF
--- a/src/mark-down/__snapshots__/markdown-formatter.spec.ts.snap
+++ b/src/mark-down/__snapshots__/markdown-formatter.spec.ts.snap
@@ -2,6 +2,8 @@
 
 exports[`MarkdownFormatter bold 1`] = `"**text**"`;
 
+exports[`MarkdownFormatter escaped 1`] = `"This string contains the \\\\<img> tag"`;
+
 exports[`MarkdownFormatter footerSeparator 1`] = `"---"`;
 
 exports[`MarkdownFormatter heading 1`] = `"## heading"`;
--- a/src/mark-down/__snapshots__/markdown-formatter.spec.ts.snap
+++ b/src/mark-down/__snapshots__/markdown-formatter.spec.ts.snap
@@ -2,8 +2,6 @@
 
 exports[`MarkdownFormatter bold 1`] = `"**text**"`;
 
-exports[`MarkdownFormatter escaped 1`] = `"This string contains the \\\\<img> tag"`;
-
 exports[`MarkdownFormatter footerSeparator 1`] = `"---"`;
 
 exports[`MarkdownFormatter heading 1`] = `"## heading"`;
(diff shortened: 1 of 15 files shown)


src/mark-down/__snapshots__/markdown-formatter.spec.ts.snap
```diff
@@ -2,6 +2,8 @@
 
 exports[`MarkdownFormatter bold 1`] = `"**text**"`;
 
+exports[`MarkdownFormatter escaped 1`] = `"This string contains the \\\\<img> tag"`;
```
karanbirsingh, line 5: This looks funky but is intentional. jest includes its own escaping in the default snapshot we create. This is partly why we use `jest-file-snapshot` for the full markdown file; otherwise, our snapshot would contain further backslashes before quotes and it would be difficult to review.

karanbirsingh: This was confusing so I got rid of the small snapshot and used `toEqual` instead.

JGibson2019 (approved):

madalynrose (approved):
>we generate a markdown snapshot file with all possible axe-core IDs and descriptions. Each time this file changes, reviewers can check whether we require changes to markdown generation.

this is so clever! awesome change!

PR #743, opened 2021-07-28
feat(ado/infra): Document release strategy

DaveTryon:
#### Details

Discuss and document how we intend to manage our releases in a way that supports multiple versions and allows users to easily control what version of the action they use. In turn, this will inform the actions taken by the build and release pipelines.

##### Motivation

Flush out assumptions

#### Pull request checklist

- [x] Addresses an existing issue: [1862304](https://dev.azure.com/mseng/1ES/_workitems/edit/1862304)
- [n/a] (after PR created) The `Accessibility Checks (pull_request)` check should fail. All other checks should pass.

### DIFF
--- a/dev/release-strategy.md
+++ b/dev/release-strategy.md
@@ -73,6 +73,7 @@ On each iteration, the ADO extension release pipeline will do the following:
 #### Pipeline questions:
 
 -   What should trigger the build pipeline? Every commit?
+-   What should trigger the release pipelines? Manually triggered?
 -   How do we tag the sources repo when the Action release pipeline runs?
 -   Are there scenarios where we would publish _just_ the action or _just_ the ADO extension?
 
--- a/dev/release-strategy.md
+++ b/dev/release-strategy.md
@@ -5,18 +5,17 @@ Licensed under the MIT License.
 
 ## Release Strategy
 
-### Repos
+### Repo
 
--   We will store the sources in the `accessibility-insights-action` repo (unchanged from today)
--   We will store the action releases in a _separate_ repo. Name is TBD, but maybe `accessibility-insights-action-dist`. For purposes of this doc, we'll call the two repos "the sources repo" and "the distribution repo"
--   For a given release (_e.g._, `v1.0.3`), we will use the same tag name in both repos to make it easy to correlate the two.
--   All changes to the distribution repo will be controlled by the pipeline
--   Each release in the distribution repo will be in a branch that matches its tag. That branch will contain just the files needed to release the action (`action.yml`, the code packed into `index.js`, etc.)
--   Consuming actions will reference the distribution repo.
+-   We will store the both the sources and the releases in the `accessibility-insights-action` repo (unchanged from today). If we later decide to change this, we will move the source code to a new repo and keep the consumption experience unchanged.
+-   All code contributions will merge into the main branch
+-   All releases will be pipeline-controlled
+-   We will create a new branch for each release of the GitHub Action. That branch will contain just the files needed to release the action (`action.yml`, the code packed into `index.js`, etc.)
+-   The existing ./dist folder of the repo will first be removed, then added to `./.gitignore`. It will exist only for local building and testing
 
-### Branches in the distribution repo
+### Release branches
 
-The distribution repo will contain 1 branch per released version. The name of the branch will be the same as the name of the release, as shown here:
+The repo will contain 1 branch per released version of the GitHub action. We are explicitly assuming that the ADO extension will release via the ADO marketplace and will not require a corresponding release branch. The name of the branch will be the same as the name of the release for the GitHub Action, as shown here:
 
 ```
 /v1.0.1
@@ -25,9 +24,11 @@ The distribution repo will contain 1 branch per released version. The name of th
 /v2.0.0
 ```
 
-### Tags in the distribution repo
+### Tags
 
-The distribution repo will have a tag that points to each specific release within the repo. In addition, each major release version will have a tag that points to the latest release of that major version. We could optionally also have a `latest` tag, which would point to the highest released version, as shown here:
+#### Release tags
+
+Each release of the GitHub Action will have a tag that points to each specific release branch within the repo. In addition, each major release version will have a tag that points to the latest release of that major version. We could optionally also have a `latest` tag, which would point to the highest released version, as shown here:
 
 ```
 /v1.0.1   <== v1.0.1
@@ -36,46 +37,59 @@ The distribution repo will have a tag that points to each specific release withi
 /v2.0.0   <== v2.0.0, v2, latest
 ```
 
+#### Source tags
+
+Each release will have a tag that points to the corresponding sources in the main branch. These tags _must_ have names that are distinct from the release tags, since they point to different branches. To make the tags as intuitive as possible, and to support releases of different packages (_e.g._, the GitHub Action vs the ADO extension), we will use the following pattern
+
+| Version | Type of release | Source tag release |
+| ------- | --------------- | ------------------ |
+| v1.0.2  | GitHub Action   | v1.0.2-sources-gh  |
+| v1.0.2  | ADO Extension   | v1.0.2-sources-ado |
+
+It is expected that there will be some cases where the `-gh` suffixed tag and the `-ado` suffixed tag will point to the same commit.
+
 ### Pipelines
 
-Logically we will have a build pipeline and 2 release pipelines (one for the GitHub action, one for the ADO extension). These may or may not be completely separate pipelines at first. Here are the responsibilities of each pipeline
+Logically we will have one build pipeline and two release pipelines (one for the GitHub action, one for the ADO extension). These may or may not be completely separate pipelines at first. Here are the responsibilities of each pipeline.
 
 #### Build pipeline
 
-On each iteration, the build pipeline will do the following:
+On each iteration, the **build pipeline** will do the following:
 
--   Clone the source repo
+-   Clone the repo (`main` branch)
 -   Build the `/dist` folder and the _signed_ ADO extension (if signing is required)
 -   Perform any possible self-validation
--   Publish the contents of the `/dist` folder as an artifact
--   Publish the ADO extension as an artifact
+-   Publish the contents of the `/dist` folder as the _Action artifact_
+-   Publish the ADO extension as the _ADO artifact_
 
 #### Action release pipeline
 
-On each iteration, the action release pipeline will do the following:
+On each iteration, the **action release pipeline** will do the following: Assume that our release version is `vX.Y.Z`
 
--   Clone the distribution repo
--   Download the action artifact created by the release pipeline
--   Create a new branch of the appropriate name (probably fail if the branch already exists)
--   Copy the action artifact into the new branch
+-   Clone the repo
+-   Download the _Action artifact_ created by the build pipeline
+-   Create a new branch (_releases/vX.Y.Z_), probably failing the release if the branch already exists
+-   Copy the _Action artifact_ into the new branch
 -   Commit the new branch
--   Create a release tag
--   Create or update the major version tag
--   Create or update the `latest` tag (if we use it)
+-   Create a release tag (_vX.Y.Z_)
+-   Create or update the major version tag (_vX_)
+-   Create or update the `latest` tag (if we use it and if no higher-versioned release already exists)
+-   Create a tag that corresponds to the SHA that build this release (_vX.Y.X-sources-gh_)
+-   Push all new tags
 
 #### ADO extension release pipeline
 
-On each iteration, the ADO extension release pipeline will do the following:
+On each iteration, the **ADO extension release** pipeline will do the following:
 
--   Download the action artifact created by the release pipeline
--   Publish the ADO artifact
+-   Download the _ADO artifact_ created by the build pipeline
+-   Publish the ADO artifact to the marketplace
+-   Create a tag that corresponds to the SHA that build this release (_vX.Y.X-sources-ado_)
+-   Push the new tag
 
 #### Pipeline questions:
 
--   What should trigger the build pipeline? Every commit?
+-   What should trigger the **build pipeline**? Every commit?
 -   What should trigger the release pipelines? Manually triggered?
--   How do we tag the sources repo when the Action release pipeline runs?
--   Are there scenarios where we would publish _just_ the action or _just_ the ADO extension?
 -   Do we need Canary/Insider/Production for release validation, or are we OK with just Production?
 
 ### Models for consuming the action
@@ -87,7 +101,7 @@ The yaml file excerpt would look something like this:
 ```
   steps:
     name: Check Accessibility
-    uses: actions/microsoft/accessibility-insights-action-dist@v1
+    uses: actions/microsoft/accessibility-insights-action@v1
 ```
 
 #### Pin to a specific release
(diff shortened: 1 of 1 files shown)


dev/release-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+Releases of other components (such as the Azure DevOps extension) will be siblings of the `/release/action` branch.
+
+The contents of each versioned branch will get completely refreshed with each new release, so syncing the version branch will always return the latest release of _that version_. We will also create a new tag (_e.g._, `v1.0.3`) with each new release. All branches under the `/release` branch will be locked down by branch policy that restricts all changes to the build pipeline. One potential exception to this practice might be a repo administrator creating the new branch (_e.g._, `/release/v2`) when a new major release is being prepared.
+
+Users wishing to always use _latest release_ of the `v1` action would include the following in their yml file:
+
+```
+  steps:
+    name: Check Accessibility
+    uses: actions/microsoft/accessibility-insights-action/release/action/v1
```
karanbirsingh, line 30: The string passed to `uses` here is a little unfamiliar - I'm used to something like [these repo docs](https://github.com/microsoft/accessibility-insights-action/blob/main/docs/usage.md#how-to-use-the-action) or these more generic docs [here](https://docs.github.com/en/actions/creating-actions/about-actions#using-release-management-for-actions). Is it intentional?

So far our docs have [recommended](https://github.com/microsoft/accessibility-insights-action/blob/main/docs/usage.md#how-to-use-the-action) tags, like `microsoft/accessibility-insights-action@v1`. Branches are nice because we don't have to update tags ourselves, but updating tags are less verbose for the end-user. What are your thoughts about branches vs tags?

dev/release-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+/release
+  /action
+    /v1
+    .
+    .
+    .
+    /vN
+```
+
+Releases of other components (such as the Azure DevOps extension) will be siblings of the `/release/action` branch.
```
karanbirsingh, line 21: I think we may not need release branches at all for the Azure DevOps extension, but I could be off; did you have something in mind?

dev/release-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+
+The contents of each versioned branch will get completely refreshed with each new release, so syncing the version branch will always return the latest release of _that version_. We will also create a new tag (_e.g._, `v1.0.3`) with each new release. All branches under the `/release` branch will be locked down by branch policy that restricts all changes to the build pipeline. One potential exception to this practice might be a repo administrator creating the new branch (_e.g._, `/release/v2`) when a new major release is being prepared.
```
karanbirsingh, line 23: I like that tags are less verbose for the user & easy for us to move around. One thing I'm unsure about is whether we anticipate creating GitHub tags for Azure Devops Extension releases. In web/unified we have tag prefixes. What are your thoughts?

If we care a lot about short tags, it might merit a separate release repository, although the consumption is more confusing then (`uses: microsoft/accessibility-insights-action-release` instead of `uses: microsoft/accessibility-insights-action`).

dev/release-strategy.md
DaveTryon: According to https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-public-action-in-a-subdirectory, it's valid to use both a branch and a tag in the `uses` block. The branch here pins the major version, and the tag pins to a specific release within that branch. There's definitely some redundancy here.

If we prefer a simpler model, we can omit the per-major-version branch and use just tags. Looking forward a bit, suppose that we have 2 major versions (1.X and 2.X) and someone wants to always use the latest version of v1. There's no v1-specific branch, so we we'd probably need to create a generic v1 tag and keep on updating it as new 1.X releases occur. It's probably better to have this complexity buried in the build pipe than having a complex consumption story.

If we don't need any releases for the ADO builds, then we could conceptually collapse everything to a single `release` branch. If/when we decide to change that, we can refactor the branches and migrate the tags. Let me take a shot at updating the doc that way...

dev/release-strategy.md
DaveTryon: I was just leaving the door open, should we need it later

dev/release-strategy.md
DaveTryon: Interesting point. In web, we use tags like `web@2.27.01 or unified@2021.0714.15`. Looking at the syntax for uses, I don't know if it would property handle something like `actions/microsoft/accessibility-insights-action@actions@1.0.3`. We could always use a character other than the @ sign, so `actions/microsoft/accessibility-insights-action@actions-1.0.3` or similar. 

Presumably, with each release, we'd also want a corresponding tag in the `main` branch, and its name would need to be different from the name that we use to release to consumers. It adds an additional wrinkle to both our consumption and pipeline stories. Let's discuss this offline to find a proposal that makes it tolerable everywhere.

DaveTryon: Updated using a separate repo for distribution, as discussed

AhmedAbdoOrtiga (approved):

JGibson2019 (commented):
LGTM!

dev/release-strategy.md
```diff
@@ -0,0 +1,111 @@
...
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+-->
+
+## Release Strategy
+
+### Repos
+
+-   We will store the sources in the `accessibility-insights-action` repo (unchanged from today)
+-   We will store the action releases in a _separate_ repo. Name is TBD, but maybe `accessibility-insights-action-dist`. For purposes of this doc, we'll call the two repos "the sources repo" and "the distribution repo"
```
AhmedAbdoOrtiga, line 11: why don't we have it in the same repo but without a dist?

AhmedAbdoOrtiga (changes requested):
.

dev/release-strategy.md
DaveTryon: @dbjorge and @peterdur, would you mind chiming in on the 1 repo or 2 repos question? We've discussed this as an engineering team and are still split. PM is fine with either approach. Here are the summary benefits of each option:

## Benefits of having the sources and the releases in _one_ repo
- Simplicity. Since code and releases all in one repo, the same GITHUB_TOKEN just works for all operations. GitHub actions in the repo have access to both source and releases.
- Symmetry. People consuming `Microsoft/accessibility-insights-action@v1.2.3` in their code know to look for the sources in the `Microsoft/accessibiiity-insights-action` repo

## Benefits of having the sources and the releases in _two different_ repos
- Size. The current `./dist` folder will go away and will basically be replicated in each release, which adds about 5MB to the repo for each release. Having 2 repos keeps the development repo nice and small.
- Consistently named tags. With 2 repos, we can have a tag like `v1.0.3` that identifies the code in the source repo, and a tag with the same name that identifies the artifact in the distribution repo. We can't do that in a single repo.
- Clean slate: We only have 1 chance to rename the distribution point, since renaming it later will impact customers who are consuming it. This is probably overengineering for what we need with this feature, but it sets us up for success down the road.

AhmedAbdoOrtiga (approved):

dev/release-strategy.md
```diff
@@ -0,0 +1,122 @@
...
+-   All releases will be workflow-controlled
+-   We will create a new branch for each release of the GitHub Action. That branch will contain just the files needed to release the action (`action.yml`, the code packed into `index.js`, etc.)
+-   The existing ./dist folder of the repo will first be removed, then added to `./.gitignore`. It will exist only for local building and testing
+
+### Release branches
+
+The repo will contain 1 branch per released version of the GitHub action. We are explicitly assuming that the ADO extension will release via the ADO marketplace and will not require a corresponding release branch. The name of the branch will be the same as the name of the release for the GitHub Action, as shown here:
+
+```
+/v1.0.1
```
dbjorge, line 21: Is the empty leading slash on these intentional? That seems prone to mistakes/confusion; I think I'd prefer to see a consistent non-empty prefix (eg, `releases/v1.0.1`). Possibly this is what you intended, based on L69?

DaveTryon: Removed

dev/release-strategy.md
```diff
@@ -0,0 +1,122 @@
...
+-   All releases will be workflow-controlled
+-   We will create a new branch for each release of the GitHub Action. That branch will contain just the files needed to release the action (`action.yml`, the code packed into `index.js`, etc.)
+-   The existing ./dist folder of the repo will first be removed, then added to `./.gitignore`. It will exist only for local building and testing
+
+### Release branches
+
+The repo will contain 1 branch per released version of the GitHub action. We are explicitly assuming that the ADO extension will release via the ADO marketplace and will not require a corresponding release branch. The name of the branch will be the same as the name of the release for the GitHub Action, as shown here:
+
+```
+v1.0.1
```
dbjorge, line 21: I strongly recommend against intentionally creating branches and tags with the same name; it makes a lot of normal-looking git commands ambiguous. I understand the desire to use short tags for usability of the action, so I'd recommend using longer branch names (eg, `releases/gh/v1.0.1`)

dev/release-strategy.md
```diff
@@ -0,0 +1,122 @@
...
+### Workflows
+
+Logically we will have one build workflow and two release workflows (one for the GitHub action, one for the ADO extension). Here are the responsibilities of each workflow.
+
+#### Build workflow
+
+This will be implemented as a GitHub action that triggers on each commit. When triggered,, the **build workflow** will do the following:
+
+-   Clone the repo (`main` branch) at the lastest SHA
+-   Build the `/dist` folder and ADO extension (Open question: Should we sign in this action?)
```
dbjorge, line 60: Even if we aren't signing here, I think we're at least going to want to be generating a NOTICE file here, which probably means this needs to be in ADO rather than a GitHub action

dev/release-strategy.md
```diff
@@ -0,0 +1,122 @@
...
+
+It is expected that there will be some cases where the `-gh` suffixed tag and the `-ado` suffixed tag will point to the same commit.
+
+### Workflows
+
+Logically we will have one build workflow and two release workflows (one for the GitHub action, one for the ADO extension). Here are the responsibilities of each workflow.
+
+#### Build workflow
+
+This will be implemented as a GitHub action that triggers on each commit. When triggered,, the **build workflow** will do the following:
```
dbjorge, line 57: nit: typo

```suggestion
This will be implemented as a GitHub action that triggers on each commit. When triggered, the **build workflow** will do the following:
```

dev/release-strategy.md
```diff
@@ -0,0 +1,122 @@
...
+-   Build the `/dist` folder and ADO extension (Open question: Should we sign in this action?)
+-   Perform any possible self-validation
+
+#### Action release workflow
+
+This will be implemented as a GitHub action that is manually triggered. When triggered, the **action release workflow** will do the following: Assume that our release version is `vX.Y.Z`
+
+-   Clone the repo (`main` branch) at a specific SHA
+-   Build the `/dist` folder as the _Action artifact_
+-   Create a new branch (_releases/vX.Y.Z_), probably failing the release if the branch already exists
```
dbjorge, line 69: Is the intention here for the new branch to be a parentless commit that contains just the built content, or to have a parent of the source SHA?

dev/release-strategy.md
DaveTryon: Fixed

dev/release-strategy.md
DaveTryon: Added to the open questions section

dev/release-strategy.md
DaveTryon: Updated

dev/release-strategy.md
dbjorge: I agree with most of the points you've written, except:

* On this "Simplicity" benefit: Because I expect us to need to use an ADO pipeline for parts of build/release (eg, for CG and NOTICE generation), I'm not sure we're really going to get a simplicity benefit here; I expect we'll still need to coordinate a token for that purpose anyway.
* I also think that "the same GITHUB_TOKEN works for all operations" is a security antipattern; I think I would prefer making it intentionally difficult to accidentally create tokens with release permissions.

Personally, my vote would go to "use action strictly as a release repo for the github action, and integrate the source into the same existing monorepo where the scan package lives, to limit the amount of cross-repo noise we have to deal with for axe-core/etc updates". I missed the team discussion about it though, so I don't want to push my opinion too strongly.

dev/release-strategy.md
```diff
@@ -0,0 +1,124 @@
...
+-->
+
+## Release Strategy
+
+### Repo
+
+-   We will store the both the sources and the releases in the `accessibility-insights-action` repo (unchanged from today). If we later decide to change this, we will move the source code to a new repo and keep the consumption experience unchanged.
+-   All code contributions will merge into the main branch
+-   All releases will be workflow-controlled
+-   We will create a new branch for each release of the GitHub Action. That branch will contain just the files needed to release the action (`action.yml`, the code packed into `index.js`, etc.)
```
AhmedAbdoOrtiga, line 13: We will create 2 things:
- GitHub Tag (release) that has the code plus the dist folder.
- Branch that's a will be used later for developers to apply fixes to this release (the branch won't have the dist)

AhmedAbdoOrtiga (changes requested):
inline

dev/release-strategy.md
dbjorge: I prefer the approach Dave originally documented, where the release branch/tag is strictly for the actually-released dist content. I don't like the idea of commit history weaving dist and non-dist commits; I would rather have a given branch be either dist-only or source-only, not weaving between both.

DaveTryon: @AhmedAbdoOrtiga, @dbjorge, after today's discussion:
- Release includes just the files needed to ship
- Release commit and tag exist outside of a branch if possible, in the `releases/gh` branch if we can't make it outside a branch
- Release workflow creates a tag (like `v1.0.3-sources-gh` to the source commit, but does not automatically create a branch. If/when we have a case where we need to create a branch, we can always create the branch on the basis of that tag.

dev/release-strategy.md
```diff
@@ -0,0 +1,118 @@
...
+-   All code contributions will merge into the main branch
+-   All releases will be workflow-controlled
+-   We will create a new commit for each release of the GitHub Action.
+-   The existing ./dist folder of the repo will first be removed, then added to `./.gitignore`. It will exist only for local building and testing
+
+### Release branches
+
+Each release will be associated with a commit containing _just the files needed for release_ (_i.e._, without the original source files). To minimize potential confusion, our preferred option is to keep each of these commits _disassociated_ from any specific branch. We believe that this is possible with relatively little effort. If, however, this proves infeasible, then we will have exactly _one_ branch used for releases of the GitHub action. That branch will be named `releases/gh`. We currently believe that the ADO extension will release via the ADO marketplace and will _not_ require a corresponding release commit. In a worst-case scenario (we need branches both for both action and for ADO), we will use a branch named `releases/ado` for ADO releases.
```
AhmedAbdoOrtiga, line 18: as per yesterday discussion with @dbjorge , I don't think we need release branches anymore, only tags

DaveTryon: @AhmedAbdoOrtiga, this was written before we knew for sure that we could skip release branches completely. I just updated it to explicitly say that there are no release branches

AhmedAbdoOrtiga (approved):

dev/release-strategy.md
```diff
@@ -0,0 +1,116 @@
...
+
+Each release will be associated with a commit containing _just the files needed for release_ (_i.e._, without the original source files). The structure of each commit will be:
+
+```
+NOTICE.html
+README.md
+action.yml
+/dist
+    index.js
+    index.js.map
```
DaveTryon, line 26: Is index.js.map really needed for consumption?

AhmedAbdoOrtiga (approved):

PR #752, opened 2021-08-03
feat(ado/infra): introduce monorepo structure using lerna

karanbirsingh:
#### Details

We plan to host the source code for both the GitHub action and an ADO extension in this repository. The two projects will share core functionality, but differ in small ways:
- different dependencies/code for pull request comments and check runs
- different metadata files (GH `action.yml` vs ADO `task.json`)
- different deployment mechanisms

Jacqueline and I are interested in hearing feedback around whether this file structure seems reasonable or whether it feels overengineered.

As a dev, it would be nice to:
- easily tell apart GH-specific files from ADO-specific files
- limit deployed dependencies to only the ones required for that product
without too much overhead or complexity.

This PR explores a lerna/yarn-workspaces monorepo similar to setups in accessibility-insights-web or accessibility-insights-service. Most of the changed files are renames (moving files from the root to `packages/gh-action/`).

Alternatives to this approach include:
- (simplest) share `package.json` for both products; create two webpack entries
- maintain separate `package.jsons` without lerna; fewer utility scripts

This PR creates the following approx file structure:
```
- lerna.json
- package.json
- yarn.lock (combined)
- common config file templates (prettier, jest, tsconfig)
- packages
  - gh-action
    - package.json
    - webpack.config.js
    - configs (prettier, jest, tsconfig)
    - src
    - dist
      - index.js / package.json / yarn.lock (copies)
  - shared
    - package.json
    - configs (prettier, jest, tsconfig)
    - src
  - ado-extension (not in PR, planned for future) 
```

Not in this PR:
- The shared package currently contains all of the source code except `index.ts` & gh-action metadata files. Eventually deps like `@octokit` will move to gh-action instead of shared, alongside client code like `pull-request-comment-creator` (`packages/ado-extension` will follow the same pattern). The IOC containers will also move to each project.
- Updating the README & docs to describe the new structure; I will do this in a subsequent PR
- Adding the ado-extension package

Drawbacks/workarounds:
- `create-image-package-metadata.js` expects externals (axe-core, etc) to be in `packages/gh-action/package.json`. Since these dependencies are in `packages/shared/package.json`, I've updated `create-image-package-metadata` to look there too. If we add these externals to `packages/gh-action/package.json` explicitly, the `package.json` files start to share many dependencies (is it still worth separating them then?)
- you have to build `shared` before building `gh-action` to pick up changes; we should look for ways to simplify this

In order to avoid duplicate versioning of dev tools (webpack, typescript, jest) I moved many devDependencies into the root folder. It means you need to run `yarn install` once at the root. The tooling will search for parent `node_modules` by default. However, if a devDependency is specific to a particular project, or if it is a CLI tool that is path-dependent, it likely needs to go in the individual project folder ([relevant issue one](https://github.com/lerna/lerna/issues/1079) and [relevant issue two](https://github.com/lerna/lerna/blob/main/README.md#common-devdependencies)). I didn't see examples of this in our project yet.

#### Pull request checklist


packages/gh-action/dist/package.json
```diff
@@ -1,5 +1,6 @@
 {
-    "name": "accessibility-insights-action",
+    "name": "accessibility-insights-gh-action",
```
DaveTryon, line 2: Is the "-gh" needed here? We'll only publish 1 action

DaveTryon: Thanks for explaining that this is just for the folder, and not reflected in any artifact

DaveTryon (approved):

.vscode/settings.json
```diff
@@ -5,15 +5,14 @@
     "prettier.requireConfig": true,
     "eslint.validate": ["typescript", "javascript"],
     "eslint.enable": true,
+    "eslint.workingDirectories": ["packages/gh-action", "packages/shared"],
```
dbjorge, line 8: [The docs for the eslint extension](https://github.com/Microsoft/vscode-eslint#settings-options) suggest that `{ "pattern": "./packages/*/" }` should work and maybe be a little less prone to future breakage

packages/shared/package.json
```diff
@@ -0,0 +1,50 @@
+{
+    "name": "shared",
```
dbjorge, line 2: I'd recommend against reusing the name of [an existing npm package](http://npmjs.com/package/shared); it's liable to cause confusing issues down the line. Consider instead using the repo name as a namespace (eg, `@accessibility-insights-action/shared`)

dbjorge:
It feels a little concerning that this increases the size of the bundled index.js by 39%; I wouldn't have expected it to change materially with this PR, so I'm suspicious there might be something missing here. Did you already investigate why the size increased so substantially? Are we bundling devDependencies accidentally?

dbjorge (changes requested):
See inline

karanbirsingh:
> It feels a little concerning that this increases the size of the bundled index.js by 39%; I wouldn't have expected it to change materially with this PR, so I'm suspicious there might be something missing here. Did you already investigate why the size increased so substantially? Are we bundling devDependencies accidentally?

@dbjorge I think this is odd but expected behavior. Here's what I found using the vs-code diff view:

`index.js` gained about 3000 lines / 200,000 characters, almost all from the [npm encoding library](https://www.npmjs.com/package/encoding). The relevant snippet is below:

before:
```js
let convert;
try {
	convert = Object(function webpackMissingModule() { var e = new Error("Cannot find module 'encoding'"); e.code = 'MODULE_NOT_FOUND'; throw e; }()).convert;
} catch (e) {}
```
after:
```js
let convert;
try {
	convert = __webpack_require__(/*! encoding */ "../../node_modules/encoding/lib/encoding.js").convert;
} catch (e) {}
```

This code [comes from an old version](https://github.com/node-fetch/node-fetch/blob/b5e2e41b2b50bf2997720d6125accaf0dd68c0ab/src/body.js#L14) of `node-fetch`. The [README at that tag](https://github.com/node-fetch/node-fetch/tree/v2.6.1#bodytextconverted) states: "(This API requires an optional dependency of the npm package encoding, which you need to install manually. webpack users may see a warning message due to this optional dependency.)"

Because `encoding` [wasn't specified in the dependencies](https://github.com/node-fetch/node-fetch/issues/412#issuecomment-379007792) of `node-fetch`, we never noticed this until now. In this PR, `encoding` is included **as a dev dependency of lerna**. webpack resolves the module and includes its source in `index.js`.

The other changes in the diff seem much smaller, so I think this is the primary cause of the change. I searched for some devDependencies like jest, and didn't see them in index.js.

AhmedAbdoOrtiga (approved):

dbjorge:
Discussed the encoding package thing with @karanbirsingh offline. He's going to look into avoiding bundling the new package (probably using `IgnorePlugin` like [this comment suggests](https://github.com/node-fetch/node-fetch/issues/412#issuecomment-379007792)). Clearing my review so it doesn't continue to block the PR after that's addressed.

PR #753, opened 2021-08-03
feat(ado/infra): Create release workflow

AhmedAbdoOrtiga:
#### Details

- Create a manual release work flow:
- The release accepts the tag and sha (if empty will use the branch)
- The release creates a tag <tag> that user consumable (has dist and needed modules).
- The release also creates a branch <releases/tag> that developers can use to continue working on this version.

##### Workflow: 
- https://github.com/AhmedAbdoOrtiga/accessibility-insights-action/actions/workflows/release.yml

##### Generated tag and branch example:
- https://github.com/AhmedAbdoOrtiga/accessibility-insights-action/releases/tag/ortiga11
- https://github.com/AhmedAbdoOrtiga/accessibility-insights-action/tree/releases/ortiga11

image

**note that manual workflows could be accessed only by users who has write access so no external user should be able to run it.**

##### Motivation

Automate the GH Action release process

#### Pull request checklist


.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+        required: true
+      commit-sha:
+        description: 'Enter the commit sha'
+        required: true     
+      
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
```
dbjorge, line 22: This needs to take `commit-sha` as an input, no?

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.OUTH }}
```
dbjorge, line 25: `OUTH` is a surprising name to me. Is it intended to be `OAUTH`? Even if so, can it use a more descriptive name instead that describes what exactly it's a token for? Naming a secret `OAUTH` is like naming a function parameter `number`.

release.sh
```diff
@@ -0,0 +1,63 @@
...
+
+echo "Setting user name and email"
+git config --global user.name "a11y-insights"
+git config --global user.email "a11y-insights-team@microsoft.com"
+
+echo "git pull"
+git pull
+
+echo "Create a new branch"
+git checkout -b releases/$tag $sha
```
DaveTryon, line 21: @dbjorge suggested that we use a slightly longer name for the branch:
```suggestion
git checkout -b releases/gh/$tag $sha
```

release.sh
```diff
@@ -0,0 +1,63 @@
...
+sed -i 's/dist/no-dist/g' .gitignore
+
+echo "git add"
+git add --all
+
+echo "git commit"
+git commit -am "commit changes"
+
+echo "push the branch"
+git push --set-upstream origin releases/$tag
```
DaveTryon, line 33: Longer branch name
```suggestion
git push --set-upstream origin releasesgh//$tag
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Build
```
dbjorge, line 35: I think ideally this would be consuming a pre-built artifact from an earlier build, rather than rebuilding at release time

dbjorge (changes requested):
See inline

DaveTryon:
@JGibson2019, what are our requirements around including 3rd party notices? This may force us to build using an ADO pipeline if this functionality is only available there

.github/workflows/release.yml
AhmedAbdoOrtiga: No, let's assume that we checout an old commit that doesn't have the release.sh file for example, it will fail the workflow itself becuase the file won't be there, here's an example:
https://github.com/AhmedAbdoOrtiga/accessibility-insights-action/runs/3244947827?check_suite_focus=true

.github/workflows/release.yml
AhmedAbdoOrtiga: Updated to 'access_token'

.github/workflows/release.yml
AhmedAbdoOrtiga: I don't think that'd be feasible in GitHub workflows scenarios, even in ADD it depends on the type of the artifacts, for example if we use a GitHub repo as an artifact, we would definitely need to build it

release.sh
```diff
@@ -0,0 +1,63 @@
...
+echo "git pull"
+git pull
+
+echo "Create a new branch"
+git checkout -b releases/$tag $sha
+
+echo "include dist to check in"
+sed -i 's/dist/no-dist/g' .gitignore
+
+echo "git add"
```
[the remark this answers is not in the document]
AhmedAbdoOrtiga: We will create 2 things (there's 2 links for an example in the description):
- GitHub Tag (release) that has the code plus the dist folder.
- Branch that's a will be used later for developers to apply fixes to this release (the branch won't have the dist)

I left a comment on Dave's PR

.github/workflows/release.yml
dbjorge: That's a good thing, not a bad thing; we *want* the version of release.sh in use to be pinned to the commit of the release. For example, if we ever release a v3 and then make some future change to the way the project is structured such that release.sh needs a corresponding change, we'd want to be able to release security patches to v2 using the v2 release.sh, not the v3 one.

.github/workflows/release.yml
dbjorge: `access_token` really isn't much better than `OAUTH`; it still doesn't tell me as a maintainer of the secrets in the repo what the access token is for.

.github/workflows/release.yml
dbjorge: SDL policy demands that the SDT tools run against the same bits that we're releasing; if we aren't able to make that happen with GitHub workflows, that probably means we need to use ADO to do the release.

.github/workflows/release.yml
DaveTryon: +1 on always using a commit sha. We haven't committed to maintaining anything like an LTS version, but it's best to make choices now that won't make it difficult if/when we eventually decide that LTS is part of our support model

release.sh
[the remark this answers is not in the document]
DaveTryon: @AhmedAbdoOrtiga, I disagree that the release should include the code. Let's talk offline

AhmedAbdoOrtiga:
Using ADO pipelines to generate third party notices file in the release bit

PR #766, opened 2021-08-09
feat(ado/infra): Creating the ado-extension package

JGibson2019:
#### Details
A follow up to #752 

This creates the ado-extension package in the monorepo, and we used the ado-prototype branch as a model. With this PR, the structure is now

```
 - packages
  - gh-action
  - shared
  - ado-extension
    - ado-extension.json
    - package.json
    - prepare-package-dir.js (a file that runs several copy operations that moves files to the dist folder)
    - task.json
    - tsconfig.json
    - jest.config.js
    - src
      - index.ts (for now, just console logs "Hello World")
      - ado-extension-test-spec.ts (Sample test file)
    - dist (ado-extension.json copy)
      - pkg (idex.js, task.json, package.json, extension-icon copies)
```
Things to note: 

- We can potentially look to use webpack instead of `prepare-package-dir.js` and this can be done in an upcoming PR
- We tested these changes using the old publisher identity (accessibility-insights-feature-crew), and we had to manually update the version number. In the future, we can look to implement automated version updating as part of the pipeline. Regardless, once we create the new publisher, we'll need to update the id values in several of the JSON files. 

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/prepare-package-dir.js
+++ b/packages/ado-extension/prepare-package-dir.js
@@ -1,3 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 // run from root folder, this scripts prepares the dist folder
 
 const fs = require('fs-extra');
--- a/packages/ado-extension/prepare-package-dir.js
+++ b/packages/ado-extension/prepare-package-dir.js
@@ -1,0 +1,42 @@
+// run from root folder, this scripts prepares the dist folder
+
+const fs = require('fs-extra');
+
+// We allow overrides of extension/task identifiers so we can deploy
+// different versions of the extension with the same YAML
+const taskJson = JSON.parse(fs.readFileSync('task.json'));
+const extensionJson = JSON.parse(fs.readFileSync('ado-extension.json'));
+
+const overrideExtensionName = process.env.ADO_EXTENSION_NAME;
+if (overrideExtensionName?.length > 0) {
+    taskJson.name = overrideExtensionName;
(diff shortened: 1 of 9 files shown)


packages/ado-extension/prepare-package-dir.js
```diff
@@ -0,0 +1,44 @@
...
+// run from root folder, this scripts prepares the dist folder
+
+const fs = require('fs-extra');
+
+// We allow overrides of extension/task identifiers so we can deploy
+// different versions of the extension with the same YAML
+const taskJson = JSON.parse(fs.readFileSync('task.json'));
+const extensionJson = JSON.parse(fs.readFileSync('ado-extension.json'));
+
+const overrideExtensionName = process.env.ADO_EXTENSION_NAME;
```
AhmedAbdoOrtiga, line 12: will this be set in the pipeline?
should it be hardcoded?

karanbirsingh: In the prototype we used variable groups in the release pipeline so that people can release to canary or test deployments. It's  very similar to the AI-service release process

AhmedAbdoOrtiga: will this be the display name for the extension in the market place? if yes then I think we need to hardcode it, in service we don't have a name for the service

AhmedAbdoOrtiga (dismissed):

AhmedAbdoOrtiga:
Just fix the build/Test, just the extension name comment but it's a non blocking

karanbirsingh (approved):

PR #771, opened 2021-08-10
feat(ado/infra): publish ado-extension distribution folder each build

karanbirsingh:
#### Details

This PR publishes `packages/ado-extension/dist` as an artifact every time the CI build runs. This allows us to create a release pipeline that consumes these artifacts and packages/publishes a VSIX file.

I limit this build to run on CI (our tests for this repository are in GitHub workflows) - doesn't make sense to create artifacts or run SDT on PRs.

I also combined the test & lint jobs for now (PublishBuildDrops is still the longest) and made sure we build before running tests (otherwise consumers of @accessibility-isnights-action/shared can't import properly).

Lastly, I add a condition to SDT so it only runs on the main branch. This lets us queue CI builds on test branches for test releases without modifying the pipeline structure too much. May come back and change this if we separate SDT into its own pipeline.

##### Motivation

enables us to get releases working - like the prototype, it'd be nice to get automated deployment early to speed up dev time

##### Context

Validation: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=25234&view=results

Not in this PR:
- refactoring gh-action drop name to avoid regressing gh-action releases; should address holistically in a separate PR
- modifying CG to produce independent NOTICE files for gh-action or ado-extension; need to address separately

#### Pull request checklist

### DIFF
--- a/pipelines/build.yaml
+++ b/pipelines/build.yaml
@@ -1,6 +1,11 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+trigger: # trigger only affects CI builds
+    branches:
+        include:
+            - main
+
 variables:
     linuxImage: 'ubuntu-18.04'
     windowsImage: 'windows-2019'
--- a/pipelines/build.yaml
+++ b/pipelines/build.yaml
@@ -31,19 +31,6 @@ jobs:
           - script: yarn lint:check
             displayName: Check for lint errors
 
-    - job: 'Tests'
-      pool:
-          vmImage: $(linuxImage)
-          demands: npm
-      steps:
-          - task: NodeTool@0
-            inputs:
-                versionSpec: '12.x'
-            displayName: use node 12.x
-
-          - script: yarn install --frozen-lockfile
-            displayName: Install dependencies
-
           - script: yarn test -- -- --ci --coverage
             displayName: Run tests
             env:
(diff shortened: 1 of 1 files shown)


pipelines/build.yaml
```diff
@@ -162,3 +155,9 @@ jobs:
                 pathtoPublish: '$(System.DefaultWorkingDirectory)/drop'
                 artifactName: 'drop'
```
AhmedAbdoOrtiga, line 156: should we update this to to something like 'gh-action-drop'

pipelines/build.yaml
```diff
@@ -53,6 +45,7 @@ jobs:
             displayName: Publish code coverage to codecov
 
     - job: 'SDT'
+      condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/main'))
```
AhmedAbdoOrtiga, line 48: will this condition make it wait for other jobs? if so that'll slow the task down but we don't need it anyway because if any job fails, we can use the whole build for release

pipelines/build.yaml
karanbirsingh: It's mentioned in the "out of scope" area, I agree it'd be good to change but I don't want to regress the gh-action release pipeline & would rather address in a separate PR

pipelines/build.yaml
karanbirsingh: Good question, It seems to evaluate the condition right away. From [these docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml) it says: "By default, steps, jobs, and stages run if all previous steps/jobs have succeeded. It's as if you specified "condition: succeeded()" (see Job status functions)." So I think because this job has no dependencies, it won't wait for anything

AhmedAbdoOrtiga (approved):

PR #798, opened 2021-08-23
feat(ado/infra): modify release pipeline to allow separate staging/prod versions

karanbirsingh:
#### Details

This PR allows us to override the staging / prod versions separately in the ADO extension release. I encountered a situation in which the staging version was x.0.1 but the production version was x.0.0 (the production release had been abandoned). In this scenario releasing to x.0.1 fails because staging already had that version (the pipeline used the same version for both).

Now only production requires explicit version numbers - staging can simply increment if not specified. It means that the version numbers for staging/production might be different; the alternative is to unlist old versions if needed, or skip version numbers if one was abandoned.

This PR also changes the GitHub releases to be drafts. This gives us time to review formatting before publishing.

##### Motivation

simplify release process

#### Pull request checklist

### DIFF
--- a/pipelines/prod-release.yaml
+++ b/pipelines/prod-release.yaml
@@ -25,9 +25,9 @@ stages:
     - stage: package_publish_staging
       variables:
           - group: ado-extension-staging
-      ${{ if ne(parameters.stagingVersionOverride, 'increment') }}:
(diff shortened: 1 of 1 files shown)


JGibson2019 (approved):
LGTM!

PR #823, opened 2021-09-03
feat(mvp-ado-extension-behavior): Add optional inputs to task.json and config

jalkire:
#### Details
This PR adds the optional `failOnAccessibilityError` and `repoServiceConnectionName` inputs to our task.json and task-config files.

##### Motivation
We'll use these in future PRs for this feature. 

##### Context
n/a

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/src/task-config/ado-task-config.spec.ts
+++ b/packages/ado-extension/src/task-config/ado-task-config.spec.ts
@@ -30,20 +30,20 @@ describe(ADOTaskConfig, () => {
     }
 
     it.each`
-        inputOption                     | inputValue        | expectedValue                                         | getInputFunc
-        ${'repoToken'}                  | ${'token'}        | ${'token'}                                            | ${() => taskConfig.getToken()}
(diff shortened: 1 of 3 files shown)


packages/ado-extension/src/task-config/ado-task-config.spec.ts
```diff
@@ -55,6 +56,21 @@ describe(ADOTaskConfig, () => {
         },
     );
 
+    it.each`
+        inputOption                   | inputValue | expectedValue | getInputFunc
+        ${'failOnAccessibilityError'} | ${true}    | ${true}       | ${() => taskConfig.getFailOnAccessibilityError()}
```
lisli1, line 61: It looks like there is only one test case in this .each test. Do you see this test having more test cases in the future?

lisli1 (approved):
Approved; one question inline

packages/ado-extension/src/task-config/ado-task-config.spec.ts
jalkire: It could, I mostly wanted this to read similarly to the above test case because they are doing almost the same thing.

PR #829, opened 2021-09-08
feat(mvp-ado-extension-behavior): enable strictNullChecks in ado-extension

lisli1:
#### Details

Enable `strictNullChecks` in ado-extension

##### Motivation

WI 1871859

##### Context

Left out enabling strictNullChecks for shared and gh-action to reduce size of PR

#### Pull request checklist

- [x] Addresses an existing issue: WI 1871859

### DIFF
--- a/packages/ado-extension/src/ado-extension.ts
+++ b/packages/ado-extension/src/ado-extension.ts
@@ -5,8 +5,10 @@ import * as adoTask from 'azure-pipelines-task-lib/task';
 
 export function runScan() {
     try {
-        const url = adoTask.getInput('url', true) || 'url';
-        console.log(`Scanning ${url}`);
+        const url = adoTask.getInput('url', true);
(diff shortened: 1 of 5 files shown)


packages/ado-extension/src/ado-extension.ts
```diff
@@ -6,7 +6,9 @@ import * as adoTask from 'azure-pipelines-task-lib/task';
 export function runScan() {
     try {
         const url = adoTask.getInput('url', true);
-        console.log(`Scanning ${url}`);
+        if (url) {
```
dbjorge, line 9: `adoTask.getInput` already guarantees that `url` will be truthy because we're passing `true` to its `required` parameter; I think it would be preferable here to instead add a `!` to the end of line 8.

AhmedAbdoOrtiga (approved):

PR #847, opened 2021-09-15
feat(mvp-ado-extension-behavior): Show previous comment

AhmedAbdoOrtiga:
#### Details

This PR updates how we show the comment in the ADO PR.
The extesnsion will:
- Create a new comment thread if none exists.
- Add a new comment for the latest run and keep the previous run details in a below comment.

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/src/progress-reporter/pull-request/ado-pull-request-comment-creator.ts
+++ b/packages/ado-extension/src/progress-reporter/pull-request/ado-pull-request-comment-creator.ts
@@ -80,36 +80,51 @@ export class AdoPullRequestCommentCreator extends ProgressReporter {
         const gitApiObject: GitApi.IGitApi = await this.connection.getGitApi();
         const prThreads = await gitApiObject.getThreads(repoId, prId);
         const existingThread = prThreads.find((p) => p.comments?.some((c) => c.content?.includes(productTitle())));
-        const existingComment = existingThread?.comments?.find((c) => c.content?.includes(productTitle()));
-        const newComment = {
+        const existingCurrentComment = existingThread?.comments?.find(
+            (c) => c.content?.includes(productTitle()) && c.content?.includes('Results from Current Run'),
+        );
+        const existingPreviousComment = existingThread?.comments?.find(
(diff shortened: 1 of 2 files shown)


packages/ado-extension/src/progress-reporter/pull-request/ado-pull-request-comment-creator.ts
```diff
@@ -101,36 +101,51 @@ export class AdoPullRequestCommentCreator extends ProgressReporter {
...
+        const existingCurrentComment = existingThread?.comments?.find(
+            (c) => c.content?.includes(productTitle()) && c.content?.includes('Results from Current Run'),
+        );
+        const existingPreviousComment = existingThread?.comments?.find(
+            (c) => c.content?.includes(productTitle()) && c.content?.includes('Results from Previous Run'),
+        );
+        const newCurrentComment = {
             parentCommentId: 0,
-            content: reportMarkdown,
+            content: '### Results from Current Run\n' + reportMarkdown,
```
jalkire, line 112: What would your thoughts be on moving this markdown-related logic over to the `ResultMarkdownConverter` and `ResultMarkdownBuilder`? Two ways to do that come to mind:
1. Add a method like `addCurrentRunTitle` that accepts the existing markdown and modifies it accordingly
2. Alternatively, add a new method like`convertWithCurrentRunTitle` that does same thing as `convert` but adds the title

and then also add a `replaceCurrentTitleWithPrevious` to modify the existing markdown. You could also export the 'Results from Current/Previous Run' strings used in `content.includes` here so they can be `const`'s living in one place. That way, all of the markdown-related logic and strings can continue to live in `ReportMarkdownConverter`/`Builder`

lisli1 (approved):

packages/ado-extension/src/progress-reporter/pull-request/ado-pull-request-comment-creator.ts
AhmedAbdoOrtiga: I agree it needs refactoring, will do it in another PR

PR #885, opened 2021-09-30
feat(baseline-implementation): Add Baseline input parameters

AhmedAbdoOrtiga:
#### Details

Add Baseline input parameter for GH action and ADO extension.

##### Motivation

Get the Action repo side ready to consume the scanner package after adding the baseline feature.

##### Context
Adding two new parameter:

1. Baseline:  Generate Baseline if true.
2. Baseline file path: Path to existing baseline file, will generate delta between the generated one and the old one if exists.

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/task.json
+++ b/packages/ado-extension/task.json
@@ -112,26 +112,11 @@
             "helpMarkDown": "Fail the task if there are accessibility issues."
         },
         {
-            "name": "baseline",
-            "type": "boolean",
-            "label": "Generate Baseline?",
-            "defaultValue": true,
-            "required": true,
-            "helpMarkDown": "Generate Baseline if true."
-        },
-        {
             "name": "baselineFile",
             "type": "string",
-            "label": "Old Baseline File Path",
+            "label": "Baseline File Path",
             "required": false,
-            "helpMarkDown": "Path to existing baseline file, will generate delta between the generated one and the old one if exists."
-        },
-        {
-            "name": "baselineName",
-            "type": "string",
-            "label": "New Baseline File Name",
-            "required": false,
-            "helpMarkDown": "The file name of the new generated baseline, newBaseline will be used if not provided and baseline is set to true."
+            "helpMarkDown": "The old baseline file path, a new baseline will be generated with the same name, if null baseline option will be disabled."
         }
     ],
     "execution": {
--- a/packages/ado-extension/task.json
+++ b/packages/ado-extension/task.json
@@ -110,6 +110,21 @@
             "defaultValue": false,
             "required": false,
             "helpMarkDown": "Fail the task if there are accessibility issues."
+        },
+        {
+            "name": "baseLine",
+            "type": "boolean",
+            "label": "Generate Baseline?",
+            "defaultValue": true,
+            "required": true,
+            "helpMarkDown": "Generate Baseline if true."
+        },
+        {
+            "name": "baseLineFile",
+            "type": "string",
+            "label": "Generate Baseline?",
+            "required": false,
+            "helpMarkDown": "Path to existing baseline file, will generate delta between the generated one and the old one if exists."
         }
     ],
     "execution": {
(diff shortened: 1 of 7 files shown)


packages/gh-action/action.yml
```diff
@@ -45,6 +45,14 @@ inputs:
...
         required: false
         default: 90000
+    base-line:
+        description: 'Generate Baseline if true.'
+        required: true
+        default: true
+    base-line-file:
+        description: 'Path to existing baseline file, will generate delta between the generated one and the old one if exists.'
+        required: true
+        default: true
```
DaveTryon, line 55: What does default of true mean for a string?

AhmedAbdoOrtiga: oops, fixed!

packages/ado-extension/task.json
```diff
@@ -110,6 +110,21 @@
...
+            "defaultValue": true,
+            "required": true,
+            "helpMarkDown": "Generate Baseline if true."
+        },
+        {
+            "name": "baseLineFile",
+            "type": "string",
+            "label": "Generate Baseline?",
+            "required": false,
+            "helpMarkDown": "Path to existing baseline file, will generate delta between the generated one and the old one if exists."
```
DaveTryon, line 127: Based on today's conversation, we'll also use the filename (without path) for the new baseline file that we create. Do we want to try to capture that in the markdown?

packages/gh-action/src/task-config/gh-task-config.ts
```diff
@@ -84,6 +84,16 @@ export class GHTaskConfig extends TaskConfig {
...
     }
 
+    public getBaseLine(): boolean {
+        return this.actionCoreObj.getBooleanInput('base-line');
+    }
+
+    public getBaseLineFile(): string | undefined {
+        const value = this.getAbsolutePath(this.actionCoreObj.getInput('base-line-file'));
+
+        return isEmpty(value) ? undefined : value;
```
DaveTryon, line 94: Somewhere along the line, we'll need code that says "if you generate a baseline file, you must provide a name". That can probably come later

packages/ado-extension/task.json
```diff
@@ -110,6 +110,21 @@
             "defaultValue": false,
             "required": false,
             "helpMarkDown": "Fail the task if there are accessibility issues."
+        },
+        {
+            "name": "baseLine",
```
DaveTryon, line 115: The 'L' should be lower case since we're using 'baseline' as a single word, not 'base line' like you'd have in baseball. That would propagate to the other places where you use spear casing (baseline-file instead of base-line-file, etc)

AhmedAbdoOrtiga: fixed!

packages/gh-action/src/task-config/gh-task-config.ts
AhmedAbdoOrtiga: fixed!

packages/ado-extension/task.json
```diff
@@ -110,6 +110,28 @@
             "defaultValue": false,
             "required": false,
             "helpMarkDown": "Fail the task if there are accessibility issues."
+        },
+        {
+            "name": "baseline",
```
dbjorge, line 115: Are all 3 of these really necessary? I would expect that `baselineFile` would be enough on its own to imply the other two. In general, I think we should avoid giving users ways of specifying input which is self-inconsistent.

packages/ado-extension/task.json
AhmedAbdoOrtiga: fixed!

packages/ado-extension/task.json
AhmedAbdoOrtiga: updated!

DaveTryon (approved):
Will the singleWorker parameter (for this repo) be a separate PR?

AhmedAbdoOrtiga:
> Will the singleWorker parameter (for this repo) be a separate PR?

yep

Issue #920, opened 2021-10-25
ADO task does not fail on a11y issues in non-PR builds


pownkel:
The ADO build task doesn't fail on accessibility errors in our manually-triggered pipeline when `failOnAccessibilityErrors` is set to `true`. @karanbirsingh pointed out that the failOnAccessibilityError function only appears in the pull request comment creator, so it is expected to also not fail on errors in CI pipelines.

#### To reproduce:
1. Create an ADO build pipeline with the ADO extension task and a url with known accessi
2. set `failOnAccessibilityError: true` and `url: https://www.washington.edu/accesscomputing/AU/before.html` (or some other page with known a11y failures)
3. Trigger the build manually. The logs will show that failures were found, but the build will not break.

dbjorge:
Coved by #942

dbjorge:
Fix released in v1.1.0

dbjorge:
Demo of the issue being fixed in a manually triggered pipeline against v1.1.0 (from the staging channel) at https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=28323&view=results (see case 7)

PR #942, opened 2021-10-30
feat(pr-baseline-2): Support workflow enforcement in GitHub repos

DaveTryon:
#### Details

The workflow enforcement code was in a class that only ran if the code was hosted in ADO, not in GitHub. This PR splits workflow-related part of the `AdoPullRequestCommentCreator` class into a new `WorkflowEnforcer` class. The IOC container is configured to create the classes so that a `WorkflowEnforcer` object is always included in the collection of `ProgressReporter` objects. If the code happens to be hosted in ADO, then we also include an `AdoPullRequestCommentCreator` before the `WorkflowEnforcer`, since we want to update the PR comment.

Since `WorkflowEnforcer` will _always_ be the last item in the chain, it's the _only_ class that should do anything that would cause the task to intentionally fail. `WorkflowEnforcer` also runs in all scenarios, regardless of the value for [Build.Reason](https://docs.microsoft.com/en-us/javascript/api/azure-devops-extension-api/buildreason)

##### Motivation

PR baselining feature

##### Context

This is a PR where I wouldn't recommend reviewing the individual commits. I took a bit of a wrong turn regarding the design then realized that it shouldn't be there.

In the interest of unblocking validation, this PR does not contain update the errors to display using the format described at https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash. We can consider that as a separate PR.

#### Pull request checklist

- [x] Addresses an existing issue: #920
- [x] (after PR created) The `Accessibility Checks (pull_request)` check should fail. All other checks should pass.

### DIFF
--- a/packages/ado-extension/src/ioc/setup-ioc-container.ts
+++ b/packages/ado-extension/src/ioc/setup-ioc-container.ts
@@ -10,7 +10,7 @@ import { AdoPullRequestCommentCreator } from '../progress-reporter/pull-request/
 import { AdoIocTypes } from './ado-ioc-types';
 import * as process from 'process';
 import { ADOArtifactsInfoProvider } from '../ado-artifacts-info-provider';
-import { WorkflowEnforcement } from '../progress-reporter/workflow/workflow-enforcement';
+import { WorkflowEnforcement } from '../progress-reporter/enforcement/workflow-enforcement';
 
 export function setupIocContainer(container = new inversify.Container({ autoBindInjectable: true })): inversify.Container {
     container = setupSharedIocContainer(container);
--- a/packages/ado-extension/src/ioc/setup-ioc-container.ts
+++ b/packages/ado-extension/src/ioc/setup-ioc-container.ts
@@ -10,6 +10,7 @@ import { AdoPullRequestCommentCreator } from '../progress-reporter/pull-request/
 import { AdoIocTypes } from './ado-ioc-types';
 import * as process from 'process';
 import { ADOArtifactsInfoProvider } from '../ado-artifacts-info-provider';
+import { WorkflowEnforcement } from '../progress-reporter/workflow/workflow-enforcement';
 
 export function setupIocContainer(container = new inversify.Container({ autoBindInjectable: true })): inversify.Container {
     container = setupSharedIocContainer(container);
@@ -17,6 +18,7 @@ export function setupIocContainer(container = new inversify.Container({ autoBind
     container.bind(AdoIocTypes.NodeApi).toConstantValue(NodeApi);
     container.bind(iocTypes.TaskConfig).to(ADOTaskConfig).inSingletonScope();
     container.bind(AdoPullRequestCommentCreator).toSelf().inSingletonScope();
+    container.bind(WorkflowEnforcement).toSelf().inSingletonScope();
     container
         .bind(iocTypes.ProgressReporters)
         .toDynamicValue((context) => {
@@ -25,11 +27,8 @@ export function setupIocContainer(container = new inversify.Container({ autoBind
                     const pullRequestCommentCreator = context.container.get(AdoPullRequestCommentCreator);
                     return [pullRequestCommentCreator];
                 }
-                case 'GitHub': {
-                    return [];
-                }
                 default: {
-                    return [];
+                    return [context.container.get(WorkflowEnforcement)];
                 }
             }
         })
--- a/packages/ado-extension/src/ioc/setup-ioc-container.ts
+++ b/packages/ado-extension/src/ioc/setup-ioc-container.ts
@@ -25,7 +25,7 @@ export function setupIocContainer(container = new inversify.Container({ autoBind
             switch (process.env.BUILD_REPOSITORY_PROVIDER) {
                 case 'TfsGit': {
                     const pullRequestCommentCreator = context.container.get(AdoPullRequestCommentCreator);
-                    return [pullRequestCommentCreator];
+                    return [pullRequestCommentCreator, context.container.get(WorkflowEnforcement)];
                 }
                 default: {
                     return [context.container.get(WorkflowEnforcement)];
--- a/packages/ado-extension/src/ioc/setup-ioc-container.ts
+++ b/packages/ado-extension/src/ioc/setup-ioc-container.ts
@@ -10,7 +10,7 @@ import { AdoPullRequestCommentCreator } from '../progress-reporter/pull-request/
 import { AdoIocTypes } from './ado-ioc-types';
 import * as process from 'process';
 import { ADOArtifactsInfoProvider } from '../ado-artifacts-info-provider';
-import { WorkflowEnforcement } from '../progress-reporter/enforcement/workflow-enforcement';
+import { WorkflowEnforcer } from '../progress-reporter/enforcement/workflow-enforcer';
 
 export function setupIocContainer(container = new inversify.Container({ autoBindInjectable: true })): inversify.Container {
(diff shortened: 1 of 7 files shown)


packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcement.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private failOnBaselineNotUpdated(suggestedBaselineUpdate: null | BaselineFileContent): void {
+        if (this.adoTaskConfig.getBaselineFile() && suggestedBaselineUpdate) {
+            throw new Error('Failed: The baseline file needs to be updated. See the PR comments for more details.');
+        }
+    }
+
+    private isSupported(): boolean {
+        return this.adoTask.getVariable('Build.Reason') == 'PullRequest';
```
lisli1, line 64: If we have this for the isSupported condition, does that mean it won't enforce workflows in CI builds?

lisli1 (approved):

jalkire (approved):
👍

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcement.ts
DaveTryon: As discussed, we'll just remove the isSupported so the enforcer always runs

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
```diff
@@ -0,0 +1,52 @@
...
+import { inject, injectable } from 'inversify';
+import { ProgressReporter } from '@accessibility-insights-action/shared';
+import { CombinedReportParameters } from 'accessibility-insights-report';
+import * as AdoTask from 'azure-pipelines-task-lib/task';
+import { BaselineEvaluation, BaselineFileContent } from 'accessibility-insights-scan';
+
+@injectable()
+export class WorkflowEnforcer extends ProgressReporter {
+    public static readonly CURRENT_COMMENT_TITLE = 'Results from Current Run';
+    public static readonly PREVIOUS_COMMENT_TITLE = 'Results from Previous Run';
```
lisli1, line 15: Just noticed this -- is this left over from copy/paste?

DaveTryon: Yes. Fixing

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public async start(): Promise<void> {
+        // We don't do anything for workflow enforcement
+    }
+
+    public async completeRun(combinedReportResult: CombinedReportParameters, baselineEvaluation?: BaselineEvaluation): Promise<void> {
+        this.failIfAccessibilityErrorExists(combinedReportResult);
+        this.failIfBaselineNeedsUpdating(baselineEvaluation);
+        return Promise.resolve();
```
dbjorge, line 28: nit: This line should be unnecessary (the `async` part of the message signature implies the same behavior if you just let the function end)

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    private failIfAccessibilityErrorExists(combinedReportResult: CombinedReportParameters): void {
+        if (this.adoTaskConfig.getFailOnAccessibilityError() && combinedReportResult.results.urlResults.failedUrls > 0) {
+            throw new Error('Failed Accessibility Error');
+        }
+    }
+
+    private failIfBaselineNeedsUpdating(baselineEvaluation?: BaselineEvaluation): void {
+        if (baselineEvaluation && this.adoTaskConfig.getBaselineFile() && baselineEvaluation.suggestedBaselineUpdate) {
+            throw new Error('Failed: The baseline file needs to be updated. See the PR comments for more details.');
```
dbjorge, line 44: This should probably be generalized a bit, since it can occur in a non-PR environment

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ProgressReporter } from '@accessibility-insights-action/shared';
+import { CombinedReportParameters } from 'accessibility-insights-report';
+import * as AdoTask from 'azure-pipelines-task-lib/task';
+import { BaselineEvaluation, BaselineFileContent } from 'accessibility-insights-scan';
+
+@injectable()
+export class WorkflowEnforcer extends ProgressReporter {
+    constructor(
+        @inject(ADOTaskConfig) private readonly adoTaskConfig: ADOTaskConfig,
+        @inject(AdoIocTypes.AdoTask) private readonly adoTask: typeof AdoTask,
```
dbjorge, line 16: This looks unused

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
DaveTryon: It was add the `Promise.resolve` or add a lint suppression. I chose the former, but I can change it

packages/ado-extension/src/ioc/setup-ioc-container.ts
```diff
@@ -10,26 +10,25 @@ import { AdoPullRequestCommentCreator } from '../progress-reporter/pull-request/
...
         .bind(iocTypes.ProgressReporters)
         .toDynamicValue((context) => {
             switch (process.env.BUILD_REPOSITORY_PROVIDER) {
                 case 'TfsGit': {
                     const pullRequestCommentCreator = context.container.get(AdoPullRequestCommentCreator);
-                    return [pullRequestCommentCreator];
-                }
-                case 'GitHub': {
-                    return [];
+                    return [pullRequestCommentCreator, context.container.get(WorkflowEnforcer)];
```
dbjorge, line 28: It feels a bit sketchy that an exception in the PR comment creator would cause the workflow enforcer not to run. I'd be okay with fixing that in a separate PR, but for now it would be nice to at least leave a comment here explaining why the order of this list is important

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+        it('succeeds in happy path (baseline not enabled)', async () => {
+            const reportStub = {} as CombinedReportParameters;
+            const baselineEvaluationStub = {} as BaselineEvaluation;
+
+            setupFailOnAccessibilityError(false);
+            setupBaselineFileParameterDoesNotExist();
+
+            workflowEnforcer = buildWorkflowEnforcerWithMocks();
+
+            await workflowEnforcer.completeRun(reportStub, baselineEvaluationStub);
```
dbjorge, line 95: I think in the "baseline not enabled" case, the expected input would be for no evaluation to be passed, not an empty evaluation

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
DaveTryon: Oops! Remnant of when we had IsSupported in this class

dbjorge (approved):

PR #964, opened 2021-11-11
chore: remove self-test validation job

karanbirsingh:
#### Details

This pull request removes the self-test validation job from pull request workflows. As described in [this changelog](https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/) Dependabot PRs respect the `permissions` keyword. We'd like to restrict dependabot PRs from having these permissions, but the self-test job becomes less useful (AI-action errors out). We're removing this validation job until it is more useful for PRs that don't have requisite permissions (perhaps a logging-only mode).

##### Motivation

Disallow Dependabot PRs from running the self-test validation workflow

##### Context

One alternative approach was conditioning the permissions on the PR author. For example, if the PR author is not dependabot, allow permissions. Since GitHub will never give write permissions to forks, this effectively leaves the core contributor team. However, any contributor rerunning the pipeline re-grants the permissions. This could be both good & bad. For now, I am removing the job entirely; we can re-add a variation soon that conditions on PR author if we'd like.

#### Pull request checklist


jalkire (approved):
Goodbye, self-test

PR #1041, opened 2022-02-09
feat(ado-improvement-1): Remove PR Comments, add output to logs

brocktaylor7:
#### Details

This removes PR comments from both the Github Action and the ADO Extension and moves the output to the logs. This output to the logs will output every time the extension runs, rather than only on PR runs.

Sample log output for Github:
```
Run echo 'Starting accessibility scanner...'
Starting accessibility scanner...
* **URLs**: 2 URL(s) failed, 2 URL(s) passed, and 0 were not scannable
* **Rules**: 6 check(s) failed, 11 check(s) passed, and 38 were not applicable
* Download the **Accessibility Insights artifact** to view the detailed results of these checks
#### Failed instances
* **3 × label**:  Ensures every form element has a label
* **2 × html-has-lang**:  Ensures every HTML document has a lang attribute
* **1 × image-alt**:  Ensures \<img> elements have alternate text or a role of none or presentation

---
This scan used [axe-core 4.3.2](https://github.com/dequelabs/axe-core/releases/tag/v4.3.2) with Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.75 Safari/537.36.
```

Sample log output for ADO:
```
##[debug] ### Results from Current Run
### ![Accessibility Insights](https://accessibilityinsights.io/img/a11yinsights-blue.svg) Accessibility Insights Action

**22 failure instances**
* (13) **label**:  Ensures every form element has a label
* (4) **color-contrast**:  Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds
* (4) **image-alt**:  Ensures \<img> elements have alternate text or a role of none or presentation
* (1) **html-has-lang**:  Ensures every HTML document has a lang attribute

**Baseline not configured**
A baseline lets you mark known failures so it's easier to identify new failures as they're introduced. See [baselining docs](https://github.com/microsoft/accessibility-insights-action/blob/main/docs/ado-extension-usage.md#using-a-baseline-file) for more.

See all failures and scan details by downloading the report from [run artifacts](undefined)

---
#### Scan summary
**URLs**: 1 with failures, 0 passed, 0 not scannable
**Rules**: 22 with failures, 11 passed, 37 not applicable

---
[...]

PR #1045, opened 2022-02-10
feat(ado-improvements-1): Remove unnecessary auth input from ADO extension

ThanyaLeif:
#### Details

This PR removes unnecessary auth input after comments in PRs were disabled from ADO Extension #1041 , since the only function of this parameter was to provide a service connection to contribute to the target repo.

Image of the inputs before this PR, including "Azure Repo Connection" field
image

Image of the inputs after this PR, "Azure Repo Connection" field was removed
image

##### Motivation

Feature work.

##### Context
The service connection is only used to grant comment writing permissions to the extension. Behavior that was remove in PR #1041 

#### Pull request checklist

- [ n/a ] Addresses an existing issue: Fixes #0000
- [n/a] (after PR created) The `Accessibility Checks (pull_request)` check should fail. All other checks should pass.


dbjorge:
Do we know whether this is a breaking change or not? ie, have you verified what happens if you have an existing build that uses this parameter, and then the extension updates to not require it anymore? If it breaks, how do we indicate a breaking change to users? If it doesn't break, should we warn users that are still passing the connection that it isn't necessary anymore?

ThanyaLeif:
> Do we know whether this is a breaking change or not? ie, have you verified what happens if you have an existing build that uses this parameter, and then the extension updates to not require it anymore? If it breaks, how do we indicate a breaking change to users? If it doesn't break, should we warn users that are still passing the connection that it isn't necessary anymore?

Yes, I verified that the old builds that use this parameter do not break after introducing the new changes.
However, users are not yet aware of this new behavior, I will discuss this with the team.

karanbirsingh:
> > Do we know whether this is a breaking change or not? ie, have you verified what happens if you have an existing build that uses this parameter, and then the extension updates to not require it anymore? If it breaks, how do we indicate a breaking change to users? If it doesn't break, should we warn users that are still passing the connection that it isn't necessary anymore?
> 
> @dbjorge Yes, I verified that the old builds that use this parameter do not break after introducing the new changes. However, users are not yet aware of this new behavior, I will discuss this with the team.

We discussed this a little further today. It seems old client usage will still work (ADO ignored the input in our test), but we agreed it's semantically a big enough change (no more comments) that bumping the major version makes sense. That way users will know to remove any service connections or manual auth they've added.

We've added an ADO workitem 1893587 that covers deprecation. This might include publishing an update to 1.x with upgrade instructions before bumping the overall version to 2.x. We propose doing that in a separate PR.

packages/ado-extension/src/task-config/ado-task-config.spec.ts
```diff
@@ -30,30 +30,29 @@ describe(ADOTaskConfig, () => {
...
-        ${'scanTimeout'}               | ${'100000'}         | ${100000}                                               | ${() => taskConfig.getScanTimeout()}
-        ${'localhostPort'}             | ${'8080'}           | ${8080}                                                 | ${() => taskConfig.getLocalhostPort()}
-        ${'localhostPort'}             | ${undefined}        | ${undefined}                                            | ${() => taskConfig.getLocalhostPort()}
-        ${'repoServiceConnectionName'} | ${'testName'}       | ${'testName'}                                           | ${() => taskConfig.getRepoServiceConnectionName()}
-        ${'baselineFile'}              | ${'./baselineFile'} | ${getPlatformAgnosticPath(__dirname + '/baselineFile')} | ${() => taskConfig.getBaselineFile()}
-        ${'failOnAccessibilityError'}  | ${true}             | ${true}                                                 | ${() => taskConfig.getFailOnAccessibilityError()}
-        ${'singleWorker'}              | ${true}             | ${true}                                                 | ${() => taskConfig.getSingleWorker()}
```
karanbirsingh, line 56: I was confused why the whole table was changed (as opposed to just the line). I realized that we are removing the longest field name. This means the whitespace for each line decreased by 1 space so the | pipe symbols are aligned.

karanbirsingh (approved):

packages/ado-extension/src/task-config/ado-task-config.spec.ts
ThanyaLeif: Yeah, this looks weird at the beginning, but as you mentioned, the formater automatically changes the arrangement of the whole table when removing the longest field name.

PR #1055, opened 2022-02-15
feat(ado-improvements-1): Add Build summary upload to ADO

brocktaylor7:
#### Details

This adds logic to upload a build summary to the ADO extension that displays in the Extension tab. You can view a sample output here: [ADO Extension run Extension Tab output](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30880&view=ms.vss-build-web.run-extensions-tab)

This also removes the PR Comment titles that were left over from when we switched things over to console.

##### Motivation

Feature work

##### Context

Currently the reportMarkdownConverter.convert() method is called in two separate places with the same parameters in the AdoConsoleCommentCreator. The LogResultsToConsole instance will change in a future PR, which is why I've left them separate instead of refactoring to only have it convert to markdown once.

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/src/output-hooks/ado-stdout-transformer.spec.ts
+++ b/packages/ado-extension/src/output-hooks/ado-stdout-transformer.spec.ts
@@ -24,10 +24,11 @@ describe(adoStdoutTransformer, () => {
         expect(output).toBe(expectedOutput);
     });
 
+    // Note, a test for ##vso[task.uploadsummary] can't be added because ADO attempts to evaluate it and fails the test suite.
+    // See results here: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30878&view=logs&j=81b7b1a0-9e2e-58a5-1601-69aaad9b82d6&t=24cd6560-9d76-5a2b-d63a-565b4fa97a0b&l=448
     it.each`
         input
         ${'##vso[task.debug]abc'}
-        ${'##vso[task.uploadsummary]abc'}
         ${'Processing page abc'}
         ${'Discovered 2 links on page abc'}
         ${'Discovered 2345 links on page abc'}
--- a/packages/ado-extension/src/output-hooks/ado-stdout-transformer.spec.ts
+++ b/packages/ado-extension/src/output-hooks/ado-stdout-transformer.spec.ts
@@ -24,8 +24,12 @@ describe(adoStdoutTransformer, () => {
         expect(output).toBe(expectedOutput);
     });
 
-    // Note, a test for ##vso[task.uploadsummary] can't be added because ADO attempts to evaluate it and fails the test suite.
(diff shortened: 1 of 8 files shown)


packages/ado-extension/src/progress-reporter/console/ado-console-comment-creator.ts
```diff
@@ -52,4 +49,28 @@ export class AdoConsoleCommentCreator extends ProgressReporter {
...
+        const reportMarkdown = this.reportMarkdownConvertor.convert(combinedReportResult, undefined, baselineInfo);
+
+        const outDirectory = this.taskConfig.getReportOutDir();
+        const fileName = `${outDirectory}/results.md`;
+
+        // eslint-disable-next-line security/detect-non-literal-fs-filename
+        if (!this.fileSystemObj.existsSync(outDirectory)) {
+            this.logger.logInfo(`Report output directory does not exist. Creating directory ${outDirectory}`);
+            // eslint-disable-next-line security/detect-non-literal-fs-filename
+            this.fileSystemObj.mkdirSync(outDirectory);
```
karanbirsingh, line 63: I'm wondering if it makes sense to create the output directory in some shared place now that multiple parts of our code reference it.

packages/ado-extension/src/output-hooks/ado-stdout-transformer.spec.ts
```diff
@@ -24,6 +24,8 @@ describe(adoStdoutTransformer, () => {
         expect(output).toBe(expectedOutput);
     });
 
+    // Note, a test for ##vso[task.uploadsummary] can't be added because ADO attempts to evaluate it and fails the test suite.
```
karanbirsingh, line 27: Is this happening because of the interpolation into `Debug tag not added - input value '$input' returned as '$input'`? If so, I'm curious about interpolating a different value rather than the raw input. It's not ideal but might solve the problem. I'm imagining using the table syntax instead, for example: https://jestjs.io/docs/api#testeachtablename-fn-timeout

brocktaylor7: I realized last night shortly after I signed off that it was just due to the test input being echoed in the output. I added a separate test with output that doesn't include the tag to address this.

packages/ado-extension/src/progress-reporter/console/ado-console-comment-creator.ts
brocktaylor7: Good call, I've moved it to `scanner.ts`. I considered moving it up a level from there to index.ts for the GH action and ado-extension.ts for the ado extension, but putting it in the scanner allows us to have it all in one place, with the taskConfig already injected to get the proper directory based on ADO vs GH. For simplicity's sake I decided to go that route.

karanbirsingh (approved):

PR #1059, opened 2022-02-18
feat(ado-improvements-1): Add console output formatting

brocktaylor7:
#### Details

This PR adds formatting for console logging. The `ResultMarkdownBuilder` is replaced with a more generic `ResultOutputBuilder` that can accept either a `MarkdownOutputFormatter` or a `ConsoleLogOutputFormatter` as a parameter, depending on which output format is desired.

This is what the console output now looks like:
```
Accessibility Insights Action

9 failure instances
* (5) label:  Ensures every form element has a label
* (2) html-has-lang:  Ensures every HTML document has a lang attribute
* (2) image-alt:  Ensures <img> elements have alternate text or a role of none or presentation

Baseline not configured
A baseline lets you mark known failures so it's easier to identify new failures as they're introduced. See baselining docs: https://github.com/microsoft/accessibility-insights-action/blob/main/docs/ado-extension-usage.md#using-a-baseline-file for more.

See all failures and scan details by downloading the report from run artifacts: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30956&view=artifacts&pathAsName=false&type=publishedArtifacts

----------------------------------------
Scan summary
URLs: 2 with failures, 2 passed, 0 not scannable
Rules: 8 with failures, 11 passed, 38 not applicable

----------------------------------------
This scan used axe-core 4.3.2: https://github.com/dequelabs/axe-core/releases/tag/v4.3.2 with Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.75 Safari/537.36.

```

This is an [example pipeline's console output](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30956&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247&l=13)

As well as the [Extensions Tab markdown output from the same run](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30956&view=ms.vss-build-web.run-extensions-tab)

##### Motivation

Feature work

##### Context

[...]

PR #1061, opened 2022-02-23
feat(ado-improvements-1): Add console output formatting and extensions tab output

brocktaylor7:
#### Details

This PR adds formatting for console logging. 

This is what the console output now looks like:
```
Accessibility Insights

22 failure instances
* (13) label:  Ensures every form element has a label
* (4) color-contrast:  Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds
* (4) image-alt:  Ensures <img> elements have alternate text or a role of none or presentation
* (1) html-has-lang:  Ensures every HTML document has a lang attribute

Baseline not configured
A baseline lets you mark known failures so it's easier to identify new failures as they're introduced. See baselining docs (https://github.com/microsoft/accessibility-insights-action/blob/main/docs/ado-extension-usage.md#using-a-baseline-file) for more.

See all failures and scan details by downloading the report from run artifacts (undefined)

-------------------
Scan summary
URLs: 1 with failures, 0 passed, 0 not scannable
Rules: 22 with failures, 11 passed, 37 not applicable

-------------------
This scan used axe-core 4.3.2 (https://github.com/dequelabs/axe-core/releases/tag/v4.3.2) with Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.75 Safari/537.36.

```

This is an [example pipeline's console output](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31056&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247&l=13)

As well as the [Extensions Tab markdown output from the same run](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31056&view=ms.vss-build-web.run-extensions-tab)

##### Motivation

Feature work

##### Context

I had initially made a generic `ReportOutputBuilder` in a previous (unmerged) PR. After talking to Karan we decided it was probably the wrong abstraction and would likely cause more problems down the road. It was easier to start from scratch and open a new PR than try to undo everything I had already done in that branch.

#### Pull request checklist


packages/shared/src/content/strings.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 export const brand = 'Accessibility Insights';
-export const title = `${brand} Action`;
```
karanbirsingh, line 5: I think this makes sense but it'd be good to audit other usages of title & its dependencies (productName, toolName) to make sure we didn't need 'Action' or 'Extension' somewhere. It seems like this was never updated to 'Extension' though for the ADO case, so it'll be good that this reference is disappearing. If we need 'Action' / 'Extension' we can move these to be product-specific.

karanbirsingh:
I think this approach seems better too. Thanks for trying it out!

This comment from the last PR about rule count still confuses me: https://github.com/microsoft/accessibility-insights-action/pull/1059#issuecomment-1048220281 reproduced below:

-----

I'm having trouble understanding the number 8 in the output. I feel like there should be 3 failed rules, not 8. Am I missing something, or does it feel wrong to you too?

```
9 failure instances
* (5) label:  Ensures every form element has a label
* (2) html-has-lang:  Ensures every HTML document has a lang attribute
* (2) image-alt:  Ensures <img> elements have alternate text or a role of none or presentation

Baseline not configured
A baseline lets you mark known failures so it's easier to identify new failures as they're introduced. See baselining docs: https://github.com/microsoft/accessibility-insights-action/blob/main/docs/ado-extension-usage.md#using-a-baseline-file for more.

See all failures and scan details by downloading the report from run artifacts: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30956&view=artifacts&pathAsName=false&type=publishedArtifacts

----------------------------------------
Scan summary
URLs: 2 with failures, 2 passed, 0 not scannable
Rules: 8 with failures, 11 passed, 38 not applicable

----------------------------------------
```

packages/ado-extension/scripts/local-overrides.json
```diff
@@ -2,5 +2,9 @@
     {
         "name": "url",
         "value": "https://markreay.github.io/AU/before.html"
+    },
```
karanbirsingh, line 5: What impact does this change have on local development? Just double-checking it's intentional, I don't know offhand how it behaves locally.

brocktaylor7: Defaulting it to false makes no difference to local development, but if you change the value to `true` it errors out with a exit code of 1.  I was using it previously for testing in local development to make sure the workflow failed in specific places. It was leftover but I decided it's not a bad thing to have easily available for development purposes. We don't have to add it in but I figured since I had used it, it might be useful to someone in the future. It wasn't immediately clear to me what the various input values were at first. I don't have a strong opinion though, if you feel like it's better to keep this file cleaner I'll happily remove it 😄

packages/shared/src/content/strings.ts
brocktaylor7: I dug into all the places that these are used. `toolName` does get passed into the `accessibility-insights-scanner` package, but is only used as metadata.  I checked with Sep as well, he said as far as output goes he's on board with the change. As far as I can tell there is nowhere that requires 'Action'/'Extension' currently.

brocktaylor7:
> I think this approach seems better too. Thanks for trying it out!
> 
> This comment from the last PR about rule count still confuses me: [#1059 (comment)](https://github.com/microsoft/accessibility-insights-action/pull/1059#issuecomment-1048220281) reproduced below:
> 
> I'm having trouble understanding the number 8 in the output. I feel like there should be 3 failed rules, not 8. Am I missing something, or does it feel wrong to you too?
> 
> ```
> 9 failure instances
> * (5) label:  Ensures every form element has a label
> * (2) html-has-lang:  Ensures every HTML document has a lang attribute
> * (2) image-alt:  Ensures  elements have alternate text or a role of none or presentation
> 
> Baseline not configured
> A baseline lets you mark known failures so it's easier to identify new failures as they're introduced. See baselining docs: https://github.com/microsoft/accessibility-insights-action/blob/main/docs/ado-extension-usage.md#using-a-baseline-file for more.
> 
> See all failures and scan details by downloading the report from run artifacts: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=30956&view=artifacts&pathAsName=false&type=publishedArtifacts
> 
> ----------------------------------------
> Scan summary
> URLs: 2 with failures, 2 passed, 0 not scannable
> Rules: 8 with failures, 11 passed, 38 not applicable
> 
> ----------------------------------------
> ```

This is being calculated in the `accessibility-insights-scan` package in `combined-report-data-converter.ts`.  It appears to be an existing bug. I'll continue digging in on this and plan to make a fix, or at least file an issue.

packages/ado-extension/scripts/local-overrides.json
karanbirsingh: Oh, I didn't notice the false. That makes sense, that way people don't need to add the key themselves!

karanbirsingh (approved):

PR #1083, opened 2022-03-08
Update ado-extension-usage documentation

mrleemurray:
Add the `scanTimeout` value unit (milliseconds) for clarity. I recently followed these instructions, encountering this issue of unknown units.

#### Details

Updating the ADO extension usage documentation to add 'milliseconds' to the `scanTimeout` optional parameter in the pipeline configuration. 

##### Motivation

I recently followed these instructions to set up AI in our pipelines & encountered the issue of not knowing whether the timeout value was in seconds or milliseconds. After some trial & error I discovered it was the latter. The update is to add my learnings so future readers don't encounter the same issue.


DaveTryon (approved):

DaveTryon:
I think you could build an argument that the parameter should be renamed, but this is a good temporary fix.

Issue #1063, opened 2022-02-23
[ADO] potential v2.x breaking changes


dbjorge:
This is a tracking issue with breaking changes that we would like to consider the next time we make an update to a new major version of the Azure DevOps extension.

* [x] Remove the Service Connection parameter (as part of dropping PR comment support) #1045 
* [x] Embed the "upload output artifact" step into the task, rather than requiring that the user add it separately #1085 
* [x] Options which users should rarely have to override should move to the bottom and be hidden by default under an "Advanced options" group which is collapsed by default. `chromePath` should go under here, and assuming we embed the "upload output artifact" step, so should `outputDir` #1092
* [x] `failOnAccessibilityError` should default to `true` instead of `false`
* [x] (see #1068) Modify the names/grouping of `siteDir`, `scanUrlRelativePath`, `localhostPort`, and `url` inputs to make it harder for users to accidentally configure mutually exclusive combinations of inputs #1122
    * Note: `inputFile`, `inputUrls`, and `discoveryPatterns` are *not* mutually exclusive with one another and should not be forced into separate groups like this. Their help text should be improved to clarify this, but improving their help text is non-breaking.

karanbirsingh:
It seems experimentally that removing the service connection parameter is non-breaking; Tania explored this and found obsolete inputs are treated as warnings.

PR #1085, opened 2022-03-10
feat(ado-ext-telemetry): Embed upload artifact into task

brocktaylor7:
#### Details

This removes the need to have a separate publish task for the report artifacts in the azure-pipelines.yml. It is using the [special logging commands](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#upload-upload-an-artifact) to upload an artifact in ADO pipelines. It also updates the documentation accordingly.

I ran pipelines for Windows, Ubuntu, and Mac to ensure operability across operating systems.  I also ran a test pipeline that still had the publish step defined in the yaml file. It fails with an error stating that the artifact already exists.

- Embed Upload Pipeline runs
	- Ubuntu Latest: [Pipelines - Run 20220310.2 logs (azure.com)](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31524&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247)
	- Windows Latest: [Pipelines - Run 20220310.4 logs (azure.com)](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31526&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247)
	- Mac Latest: [Pipelines - Run 20220310.6 logs (azure.com)](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31528&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247)
	- What happens if pipeline still has publish step for artifacts (it fails that step): [Pipelines - Run 20220310.8 logs (azure.com)](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31554&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=a6453dc2-c13f-50dd-7490-386079329673)

##### Motivation

Feature work, partially addressing #1063

##### Context

I initially attempted to use the `uploadArtifact` method defined in the `azure-pipelines-task-lib/task` library only to find out that under the hood it uses the special logging command via a `console.log` call.  We are capturing and modifying the `stdout`, and anything that uses a console.log receives a `##debug` prefix.  Once I realized that the underlying mechanism was to use the logging commands, I switched to using the logging command directly.

#### Pull request checklist

### DIFF
--- a/docs/ado-extension-usage.md
+++ b/docs/ado-extension-usage.md
@@ -147,7 +147,7 @@ You can choose to block pull requests if the extension finds accessibility issue
 
 ## Running multiple times in a single pipeline
 
-If you want to run the extension multiple times in a single pipeline, you will need to ensure that a unique `artifactName` input is specified in your YAML file for each task. Artifact names must be unique across all tasks in a pipeline.
+If you want to run the extension multiple times in a single pipeline, you will need to ensure that unique `artifactName` and `outputDir` inputs are specified in your YAML file for each task. Artifact names and the output directory must be unique across all tasks in a pipeline.
 
 ## Troubleshooting
 
--- a/docs/ado-extension-usage.md
+++ b/docs/ado-extension-usage.md
@@ -133,10 +133,15 @@ Here is an example of a YAML file that is configured to take advantage of a base
       baselineFile: '$(Build.SourcesDirectory)/baselines/my-web-site.baseline'
 ```
 
-## Viewing results
+## Report Artifacts
 
--   Summary results along with a link to report artifacts are output in both the task log and in the Extensions tab of the pipeline run. To view the task log, click into the job and then click on the accessibilityinsights task.
--   An HTML report containing detailed results is saved as a pipeline artifact. To view it, navigate to Artifacts from the build. Under `accessibility-reports`, you'll find the downloadable report labeled `index.html`.
+By default, an HTML report containing detailed results is automatically uploaded as a pipeline artifact named `accessibility-reports`. You can opt out of this automatic artifact upload by setting the `uploadResultAsArtifact` parameter to `false`. You can also specify a custom artifact name by setting the `artifactName` parameter in your YAML file. If not opted out, a link to the artifacts will also appear in both the task log and in the Extensions tab of the pipeline run.
+
+To view the report, navigate to Artifacts from the build. Under `accessibility-reports`, or the artifact name manually specified, you'll find the downloadable report labeled `index.html`.
+
+## Summary results
+
+-   Summary results are output in both the task log and in the Extensions tab of the pipeline run. To view the task log, click into the job and then click on the `accessibilityinsights` task.
 
 ## Blocking pull requests
 
--- a/docs/ado-extension-usage.md
+++ b/docs/ado-extension-usage.md
@@ -37,11 +37,6 @@ steps:
           # Provide either siteDir or url
           # siteDir: '$(System.DefaultWorkingDirectory)/path-to-built-website/'
           # url: 'your-website-url'
-
-    - publish: '$(System.DefaultWorkingDirectory)/_accessibility-reports'
-      displayName: Upload report artifact
-      condition: succeededOrFailed()
-      artifact: 'accessibility-reports'
 ```
 
 ### Scan a URL
@@ -136,18 +131,12 @@ Here is an example of a YAML file that is configured to take advantage of a base
   inputs:
       url: 'http://localhost:12345/'
       baselineFile: '$(Build.SourcesDirectory)/baselines/my-web-site.baseline'
-
-- publish: '$(System.DefaultWorkingDirectory)/_accessibility-reports'
-  displayName: Upload report artifacts
-  condition: succeededOrFailed()
-  artifact: 'accessibility-reports'
 ```
 
 ## Viewing results
 
--   An HTML report containing detailed results is saved to disk. To view it, you need to have added the step to upload artifacts to your yml file (see [Basic template](#basic-template)). Navigate to Artifacts from the build. Under `accessibility-reports`, you'll find the downloadable report labeled `index.html`.
-
--   If the workflow was triggered by a pull request, the action should leave a comment on the Azure DevOps pull request with results. The extension does not leave comments on repos in GitHub.
+-   Summary results along with a link to report artifacts are output in both the task log and in the Extensions tab of the pipeline run. To view the task log, click into the job and then click on the accessibilityinsights task.
+-   An HTML report containing detailed results is saved as a pipeline artifact. To view it, navigate to Artifacts from the build. Under `accessibility-reports`, you'll find the downloadable report labeled `index.html`.
 
 ## Blocking pull requests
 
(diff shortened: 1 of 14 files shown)


ThanyaLeif (dismissed):
Nicely done!

packages/ado-extension/src/progress-reporter/console/ado-console-comment-creator.ts
```diff
@@ -62,6 +63,11 @@ export class AdoConsoleCommentCreator extends ProgressReporter {
         this.logger.logInfo(`##vso[task.uploadsummary]${fileName}`);
     }
 
+    private uploadReportArtifacts(): void {
+        const workingDirectory = this.taskConfig.getVariable('System.DefaultWorkingDirectory') ?? '';
+        this.logger.logInfo(`##vso[artifact.upload artifactname=accessibility-reports]${workingDirectory}/_accessibility-reports`);
```
dbjorge, line 68: I think the `_accessibility-reports` part can be overriden by one of the task inputs, no?

brocktaylor7: Great catch! I've updated this to use the `outputDir` input.

packages/ado-extension/src/progress-reporter/console/ado-console-comment-creator.ts
```diff
@@ -62,6 +63,12 @@ export class AdoConsoleCommentCreator extends ProgressReporter {
         this.logger.logInfo(`##vso[task.uploadsummary]${fileName}`);
     }
 
+    private uploadReportArtifacts(): void {
+        const workingDirectory = this.taskConfig.getVariable('System.DefaultWorkingDirectory') ?? '';
+        const outputDirectory = this.taskConfig.getReportOutDir();
+        this.logger.logInfo(`##vso[artifact.upload artifactname=accessibility-reports]${workingDirectory}/${outputDirectory}`);
```
karanbirsingh, line 69: what happens if someone includes our task twice in their pipeline - does it overwrite the old file? if so, I wonder if we should be resilient to that or if it's OK to require the user to pick separate names

brocktaylor7: That's an interesting point! There are two considerations here:

1. Whether the files from task 2 overwrite the files from task 1 in the working directory. I'll need to investigate this further, but we could probably clear the working directory at the end of the task after the artifact upload, which would leave a clean working directory for the next task. (It feels like it's possible this happens automatically between tasks?)
2. Whether the artifacts from task 2 overwrite the artifacts from task 1. We already know that it doesn't overwrite the existing artifacts, it would fail due to the artifact of the same name already existing as it did here: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31554&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=a6453dc2-c13f-50dd-7490-386079329673&l=31 With that being the case, maybe it makes sense to have an input for the `artifactname` instead of it being hard-coded to `accessibility-reports`. That way you could have a separate artifact name in the two tasks. This would require the user to configure the tasks with separate artifact names for it to work.  I'm not sure there's any clean way around that.

karanbirsingh: That's true, the same scenario applies to the locally saved output. IMO the immediate issue is we shouldn't crash if run twice; I don't have a strong preference between options to remediate it, e.g. accepting an `artifactName` and/or creating unique names ourselves.

brocktaylor7: I've added an artifactName input that allows for the task to be run multiple times in the same pipeline. 

I also made it resilient to the job being run multiple times. [Pipeline artifacts of the same name can only be published once per build](https://github.com/MicrosoftDocs/azure-devops-docs/issues/10543) but there is a system variable that returns the number of job attempts, this allows us to add a suffix to the artifactName if it's a second job attempt.  With this workaround it no longer fails for duplicate artifacts in a second job attempt. 

To demonstrate, I created a [test pipeline that runs the task twice](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=31631&view=logs&j=6b902995-b73d-5f5c-66fd-a7f66c857d2c&t=0bba9c5b-99e1-55c0-914e-3a2bee3e6d65), and I also ran the job twice. It uploads artifacts for each job run, and only fails due to the `failOnAccessibilityError` being set to true, which is expected behavior. 

I've also added documentation to outline the necessary steps for this scenario.

Let me know if there is anything that you feel I've missed!

packages/ado-extension/task.json
```diff
@@ -118,6 +118,14 @@
...
             "defaultValue": true,
             "helpMarkDown": "To get deterministic scanning results, either specify the singleWorker parameter or ensure that the value specified for the maxUrls parameter is larger than the total number of urls in the web site being scanned."
+        },
+        {
+            "name": "artifactName",
+            "type": "string",
+            "label": "Artifact Name",
+            "required": false,
+            "defaultValue": "accessibility-reports",
+            "helpMarkDown": "Name of the report artifact to be uploaded to the build."
```
dbjorge, line 128: Some Pipelines environments (notably, OneBranch) forbid individual tasks like ours from uploading artifacts directly. In OneBranch, for example, the usual usage is for individual tasks to emit output into a specific standard output directory, and for all artifacts to go through an assortment of standardized scans/validation before being uploaded by a final centralized upload task.

I think the most straightforward way to enable this use case would be:

* Keep the `required: false` and non-empty `defaultValue` here as-is
* Update `helpMarkdown` appending something like `Leave this blank to skip uploading the output as an artifact."
* Update the impl to skip the artifact upload step if it's blank

dbjorge: An explicit "Upload results as an artifact?" boolean would also be fine instead of allowing blank values, and maybe more consistent with the learning we're just now applying regarding user confusion around some of the other parameters with special behavior when left blank

packages/ado-extension/src/progress-reporter/console/ado-console-comment-creator.ts
karanbirsingh: Thanks, that helps!

docs/ado-extension-usage.md
```diff
@@ -136,30 +131,26 @@ Here is an example of a YAML file that is configured to take advantage of a base
...
-  condition: succeededOrFailed()
-  artifact: 'accessibility-reports'
 ```
 
 ## Viewing results
 
--   An HTML report containing detailed results is saved to disk. To view it, you need to have added the step to upload artifacts to your yml file (see [Basic template](#basic-template)). Navigate to Artifacts from the build. Under `accessibility-reports`, you'll find the downloadable report labeled `index.html`.
-
--   If the workflow was triggered by a pull request, the action should leave a comment on the Azure DevOps pull request with results. The extension does not leave comments on repos in GitHub.
+-   Summary results along with a link to report artifacts are output in both the task log and in the Extensions tab of the pipeline run. To view the task log, click into the job and then click on the accessibilityinsights task.
```
dbjorge, line 138: This maybe needs an update now that uploading the artifact is optional

packages/ado-extension/task.json
```diff
@@ -118,6 +118,22 @@
...
+            "required": true,
+            "helpMarkDown": "Automatically upload the result as an artifact to the build. Set to false if you need to upload the artifact manually in a separate task or publish step."
+        },
+        {
+            "name": "artifactName",
+            "type": "string",
+            "label": "Artifact Name",
+            "required": false,
+            "defaultValue": "accessibility-reports",
+            "helpMarkDown": "Name of the report artifact to be uploaded to the build."
```
dbjorge, line 136: nit: It might be good to be explicit about how this interacts with `uploadResultsAsArtifact` by adding a `visibleRule` for pipeline UX users and a `Ignored if uploadResultAsArtifact is false.` to the `helpMarkdown` for yaml users

dbjorge (approved):
LGTM modulo the minor suggestion for the `artifactName` settings in `task.json`

PR #1088, opened 2022-03-12
feat(ado-ext-telemetry): Making failOnAccessibilityError true by default

ThanyaLeif:
#### Details

This PR makes `failOnAccessibilityError` `true` by default, to encourage our users to resolve accessibility issues detected by our extension.

##### Motivation

Feature work, partially addressing [#1063](https://github.com/microsoft/accessibility-insights-action/issues/1063)

##### Context

Warnings are commonly ignored since they do not block PRs, making failOnAccessibilityError true by default will block PRs with accessibility issues, rather than leave it up to the user.

The user can change this property again but must do so explicitly during pipeline configuration.

I considered deleting step 2 in the "Blocking pull request" section in the `ado-extension-usage` file now that `failOnAccessibilityError` is set to `true` by default, but since the property can still be modified by the user I decided to keep it.

#### Pull request checklist


brocktaylor7:
In reference to your context about `ado-extension-usage.md`, I think it probably makes sense to update the documentation to note that the default is that it blocks, and you can choose to not block by setting it to false.

ThanyaLeif:
> In reference to your context about `ado-extension-usage.md`, I think it probably makes sense to update the documentation to note that the default is that it blocks, and you can choose to not block by setting it to false.

Done!

brocktaylor7 (approved):
LGTM!

PR #1093, opened 2022-03-21
feat(ado-ext-telemetry): update canary release pipeline to allow an extension version override

brocktaylor7:
#### Details

This allows for the canary release pipeline version to be overridden.  Before this, even if the `versionOverride` parameter was set, it would look at the `ado-extension-canary` variable group and throw it out.  This allows us to upgrade the canary version by overriding the default 'ignore-if-canary' parameter in the build.

We need this to be able to upgrade the major version of Canary as we prepare to merge in breaking changes.

##### Motivation

Feature work

#### Pull request checklist

### DIFF
--- a/pipelines/canary-release.yaml
+++ b/pipelines/canary-release.yaml
@@ -12,7 +12,7 @@ parameters:
       default: 'ignore-if-canary'
 
 variables:
-    - ${{ if ne(parameters.variableGroupName, 'ado-extension-canary') }}:
+    - ${{ if and(ne(parameters.variableGroupName, 'ado-extension-canary'), ne(parameters.versionOverride, 'ignore-if-canary')) }}:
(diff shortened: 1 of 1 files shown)


pipelines/canary-release.yaml
```diff
@@ -12,7 +12,7 @@ parameters:
       default: 'ignore-if-canary'
 
 variables:
-    - ${{ if ne(parameters.variableGroupName, 'ado-extension-canary') }}:
+    - ${{ if and(ne(parameters.variableGroupName, 'ado-extension-canary'), ne(parameters.versionOverride, 'ignore-if-canary')) }}:
```
dbjorge, line 15: Should this be `or` rather than `and`?

dbjorge (approved):

PR #1107, opened 2022-03-25
feat(ado-ext-telemetry-1893588): pass release metadata to the ADO extension

dbjorge:
#### Details

This PR enables the running ADO extension to query certain metadata about itself that gets filled in at release time. In particular, this allows us to embed a release-specific Application Insights Connection String, as well as the specific version/environment being used (so we can include that in telemetry).

I've made sure that it supports a null connection string for the sake of any future public releases, where we won't want to include telemetry.

##### Motivation

See feat(ado-ext-telemetry-1893588)

##### Context

The next PR will read this metadata as part of initializing an Application Insights telemetry client.

#### Pull request checklist

- [x] Addresses an existing issue: 1930556
- [x] Added relevant unit test for your changes. (`yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`yarn precheckin`)
- [ ] (after PR created) The `Accessibility Checks (pull_request)` check should fail. All other checks should pass.

PR #1116, opened 2022-04-01
feat(ado-ext-telemetry-1893588): implement App Insights telemetry entry point

dbjorge:
#### Details

This PR creates an initial implementation of App Insights telemetry based on the `AdoExtensionMetadata` added in #1107. The major pieces are:

* The `shared` package contains a `TelemetryClient` interface which can track `TelemetryEvent`s. It contains a `NullTelemetryClient` which no-ops telemetry events by default. The GitHub Action does not override this behavior, so it always no-ops telemetry.
* The `shared` `scanner.ts` implementation is updated to send a single `ScanStart` event with no properties. A future PR will add a `ScanEnd` event that includes properties about number of failures/baselining state/etc
* The `ado-extension` adds a reference to the `accessibilityinsights` package and implements an `AppInsightsTelemetryClient` based on it. I have set up the client such that it omits all autocollected data not requested by the feature spec, and to include as `commonProperties` all ADO environment properties requested by the spec, plus some data from the `AdoExtensionMetadata` so we can distinguish which environment telemetry is coming from
* The `ado-extension` adds a `TelemetryClientFactory` which decides whether to use an `AppInsightsTelemetryClient` or a `NullTelemetryClient` based on the presence of an `appInsightsConnectionString` property in the release environment's metadata. The intention is that it will only use App Insights telemetry in release environments whose ADO release variable groups explicitly contain an app insights connection string; this should be omitted in any future public releases.
* I've updated the `run-locally.js` tool to support the new `AdoExtensionMetadata` stuff - this can be used for testing telemetry locally.

I've used the run-locally support to validate that telemetry makes its way to our new dev app insights instance; team members should have permissions to view the dev app insights instance already and can find instructions by looking up `a11y-insights-action-dev-telemetry-app-insights` in our team OneNote.

##### Motivation

Feature work (feature 1893588 / story 1930551)

##### Context

Out of scope: a future PR will be finalizing the specific data (# failures, baseline info, etc) that we log from the scanner

#### Pull request checklist

- [x] Addresses an existing issue: story 1930551

### DIFF
--- a/packages/ado-extension/package.json
+++ b/packages/ado-extension/package.json
@@ -33,7 +33,7 @@
     },
     "dependencies": {
         "@accessibility-insights-action/shared": "^1.0.0",
-        "@microsoft/applicationinsights-web-basic": "^2.7.4",
+        "applicationinsights": "^2.2.2",
         "azure-pipelines-task-lib": "^3.2.0",
         "reflect-metadata": "^0.1.13"
     }
--- a/packages/ado-extension/package.json
+++ b/packages/ado-extension/package.json
@@ -33,7 +33,7 @@
     },
     "dependencies": {
         "@accessibility-insights-action/shared": "^1.0.0",
-        "@microsoft/applicationinsights-web": "^2.7.4",
+        "@microsoft/applicationinsights-web-basic": "^2.7.4",
         "azure-pipelines-task-lib": "^3.2.0",
         "reflect-metadata": "^0.1.13"
     }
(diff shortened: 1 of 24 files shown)


brocktaylor7 (approved):
This looks great! I especially appreciated the clarifying comments on `app-insights-telemetry-client.ts`. I was looking through the docs on the library and saw that the default implementation uses `setup()`, your comments made the reasoning behind your implementation very clear and will prevent problems from cropping up in the future.  LGTM!

PR #1122, opened 2022-04-04
feat(ado-ext-telemetry): Avoiding configuration of mutually exclusive parameters

### LINKED ISSUE #1068: It needs to be more obvious that siteDir and url are mutually exclusive inputs
## Describe the bug

Several pilot teams onboarding to the Azure DevOps task have gotten confused about exactly which parameters they should be specifying when setting up the task. In particular, we've seen several pilot teams try to specify both `siteDir` and `url` inputs simultaneously, even though these parameters are mutually exclusive (specifying `url` causes `siteDir`, `localhostPort`, and `scanUrlRelativePath` to be ignored).

The task should do a better job of making it hard for a user to make this mistake. For users of ADO's graphical Pipeline definition UX (most of our pilot teams so far), I think the ideal would be to use task definition groups to enforce the parameters' mutually exclusivity as follows:

* Add a new required option that the user must select to determine whether they want to use "our task hosts your site for you using a simple Express configuration" mode vs "you host your site in a separate task before this one runs" mode
    * siteDir, scanUrlRelativePath, and localhostPort should be moved to a group with is only visible when "we host your site for you" mode is selected. In this mode, siteDir should be required and the others should be optional. The guidance for each should clarify the exact expected format (particularly, siteDir should clarify if it accepts relative vs absolute paths).
    * url should only be visible (and should be required) when "you host your site separately" mode is selected

However, this wouldn't do anything to help users of yaml build definitions. For them, I'd suggest the following additions:

* add an early parameter validation step before starting to scan which detects mutually exclusive parameters being set and fails fast with an actionable error message, rather than silently picking one set to ignore
* Change the names of the inputs to make the grouping between them more clear and to make it easier for documentation to refer to the parameters as a group (eg, for error messages and the `url` documentation to explain what it's mutually exclusive with). Perhaps something like:
    * `staticSiteDir` (instead of `siteDir`)
    * `staticSitePort` (instead of `localhostPort`)
    * `staticSiteUrlRelativePath` (instead of `scanUrlRelativePath`)

Note that most of the above suggestions would be breaking changes; see #1063 

**To Reproduce**

1. create a test pipeline using the ADO task which specifies both `siteDir` and `url`
2. run a build
3. observe that the task silently ignores `siteDir`

## CodePen repro example

n/a

## Expected behavior

We should strive for our UX to be a "pit of success" - we should make it hard for users to misconfigure the product, especially in a way which misbehaves silently. It isn't enough to rely on users reading usage documentation.

## Screenshots

n/a

- OS Name & Version: n/a
- AI-Action Version & Environment: v1.1.1 production
- Browser Version: n/a
- Target Page: n/a

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

yes

## Additional context

n/a

ThanyaLeif:
#### Details

The purpose of this PR is to make it more apparent which inputs are mutually exclusive when configured. 

> Since several pilot teams onboarding to the Azure DevOps task have gotten confused about exactly which parameters they should be specifying when setting up the task. #1068 

As suggested by bug #1068 two approaches were taken  one for the UX users, and other one for YAML users.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              

For users of ADO's graphical Pipeline definition UX I added a visibility rule that groups the `staticSiteDir`, `staticSitePort` and `staticUrlRelativePath` inputs in the new `staticSite` mode ("You host your own site" option suggested by #1068). And the `url` input in the `dynamicSite` mode.

Both modes are mutually exclusive, which ensures that one group of options is not shown when the other has been selected.

For users of yaml build definitions, parameter validations have been added to ensure that inputs of the above modes are not set at the same time, especially `url` and` staticSiteDir`.

Image of the static SiteDir, staticSitePort and static SiteUrl RelativePath inputs shown in the graphical interface when the "We host your site" mode is selected
<img width="258" alt="image" src="https://user-images.githubusercontent.com/32908302/161642854-c659e43c-3deb-4e2d-bbdc-0240a32cdcf5.png">


Image of the inputs url shown in the graphical interface when the "Host your own site" mode is selected
<img width="252" alt="image" src="https://user-images.githubusercontent.com/32908302/161643199-1ca11544-4a46-404a-a4c3-cbf0203e0a85.png"> 

Error message when an input is configured in the wrong mode
<img width="703" alt="image" src="https://user-images.githubusercontent.com/32908302/161643657-a45154e0-d7b0-424f-a304-6f102edd13e7.png">


##### Motivation

Address issue #1068 and partially address issue #1063 

##### Context
For the new `hostingMode` input, the `required` parameter only affects the inputs in the classic view of the pipeline (the UI).
I intentionally left out validating that parameter for YAML users because its function is solely to show and hide the required parameters for each mode in the UX.
But I added an extra validation in `workflow-enforcer.ts` `line 19`, which ensures that mutually exclusive parameters are not set at the same time even if the hosting mode is not specified.
I did some tests consuming my pipeline and verified that hostingMode can be left out of configuration in YAML builds without affecting the workflow.

#### Pull request checklist
<!-- If a checklist item is not applicable to this change, write "n/a" in the checkbox -->
- [x] Addresses an existing issue: Fixes #1068 


packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
```diff
@@ -35,6 +39,67 @@ export class WorkflowEnforcer extends ProgressReporter {
...
+                `A configuration error has occurred, Url and staticSiteDir inputs cannot be set at the same time`,
+            );
+            await this.failRun();
+            return true;
+        }
+        return false;
+    }
+
+    private async failIfUrlIsConfiguredInStaticSiteMode(): Promise<boolean> {
+        if (this.adoTaskConfig.getHostingMode() === 'staticSite' && typeof this.adoTaskConfig.getUrl() !== 'undefined') {
```
brocktaylor7, line 56: nit: is there a reason to check the type rather than just checking if it's undefined? In the above function you check if `!== undefined` and this and the two below it you check if the `typeof` the variable `!== 'undefined'`. I prefer the former, but am okay with either so long as they stay consistent :)

packages/ado-extension/task.json
```diff
@@ -25,35 +25,54 @@
             "groupName": "advancedOptions"
         },
         {
-            "name": "siteDir",
+            "name": "hostingMode",
+            "type": "pickList",
+            "label": "Hosting Mode",
+            "required": true,
```
brocktaylor7, line 31: With hosting mode being a required input, we should probably make changes to the docs to specify this, including adding it to relevant example YAML snippets in the docs.

brocktaylor7 (changes requested):
Looks good! Just a few small suggestions :)

packages/ado-extension/task.json
brocktaylor7: After chatting it appears that the required is only pertinent to the classic UI for configuration within ADO and the YAML files don't require it, so we're good on this front.

brocktaylor7 (approved):

packages/ado-extension/task.json
ThanyaLeif: From my understanding, the "required" parameter only affects the inputs in the classic view of the pipeline (the UI).
I intentionally left out validating that parameter for YAML users because its function is solely to show and hide the required parameters for each mode in the UX.
But I added an extra validation in `workflow-enforcer.ts` `line 19`, which ensures that mutually exclusive parameters are not set at the same time even if the hosting mode is not specified.
I did some tests consuming my pipeline and verified that hostingMode can be left out of configuration in YAML builds without affecting the workflow.

packages/ado-extension/src/progress-reporter/enforcement/workflow-enforcer.ts
ThanyaLeif: Fixed :)

PR #1126, opened 2022-04-07
feat(ado-ext-telemetry): add telemetry sender progress reporter

brocktaylor7:
#### Details

This introduces a new progress reporter in the ADO Extension that collects the relevant data points and tracks the event via the `TelemetryClient`. This also changes `TelemetryEvent.properties` to accept property values of `any` type instead of strictly `string`. This makes it easier for us to pass data of all kinds. I verified that this is supported by the ApplicationInsights API based off the documentation found here: https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#properties

##### Motivation

Feature work

##### Context

We considered putting this logic in `scanner.ts` in the `shared` package, but it seemed like it would be the wrong abstraction as the data points needed and where to get them will likely differ between the Github Action and the ADO Extension. 

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
+++ b/packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
@@ -27,7 +27,7 @@ export class TelemetrySender extends ProgressReporter {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const eventProperties: { [key: string]: any } = {};
         eventProperties.teamProject = this.adoTaskConfig.getTeamProject() ?? '';
-        eventProperties.runId = this.adoTaskConfig.getRunId()?.toString() ?? '';
+        eventProperties.runId = this.adoTaskConfig.getRunId() ?? '';
 
         eventProperties.rulesFailedListWithCounts = combinedReportResult.results.resultsByRule.failed.map((failuresGroup) => {
             const failureCount = failuresGroup.failed.reduce((a, b) => a + (b.urls ? b.urls.length : 0), 0);
@@ -44,7 +44,7 @@ export class TelemetrySender extends ProgressReporter {
         }
 
         this.telemetryClient.trackEvent({
-            name: 'ScanStart',
+            name: 'ScanCompleted',
             properties: eventProperties,
         } as TelemetryEvent);
     }
--- a/packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
+++ b/packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
@@ -9,8 +9,6 @@ import { BaselineEvaluation } from 'accessibility-insights-scan';
 
 @injectable()
 export class TelemetrySender extends ProgressReporter {
-    private scanSucceeded = true;
-
     constructor(
         @inject(ADOTaskConfig) private readonly adoTaskConfig: ADOTaskConfig,
         @inject(iocTypes.TelemetryClient) private readonly telemetryClient: TelemetryClient,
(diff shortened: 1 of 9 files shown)


packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    public async start(): Promise<void> {
+        // We don't send anything for telemetry
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public async completeRun(combinedReportResult: CombinedReportParameters, baselineEvaluation?: BaselineEvaluation): Promise<void> {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const eventProperties: { [key: string]: any } = {};
+        eventProperties.teamProject = this.adoTaskConfig.getTeamProject() ?? '';
```
dbjorge, line 29: These two should already be getting populated as commonProperties in the client constructor, no? Why repeat them here?

packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
```diff
@@ -0,0 +1,59 @@
...
+        eventProperties.runId = this.adoTaskConfig.getRunId() ?? '';
+
+        eventProperties.rulesFailedListWithCounts = combinedReportResult.results.resultsByRule.failed.map((failuresGroup) => {
+            const failureCount = failuresGroup.failed.reduce((a, b) => a + (b.urls ? b.urls.length : 0), 0);
```
ThanyaLeif, line 33: Not completely understanding how this function works.
What 'a' and 'b' stands for in the failureCount reduce function?

packages/ado-extension/src/telemetry/app-insights-telemetry-client.ts
```diff
@@ -52,7 +52,6 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
     }
 
     public async flush(): Promise<void> {
-        this.logger.logDebug(`[Telemetry] flushing telemetry`);
```
ThanyaLeif, line 55: What was this entry used for?

ThanyaLeif (commented):
I just added a couple of comments for clarification purposes, but the code looks good in general. :)

packages/ado-extension/src/telemetry/app-insights-telemetry-client.ts
brocktaylor7: This was used to output to the console that it was flushing telemetry. I'm okay with leaving it in if anyone has strong feelings on it!

packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
```diff
@@ -0,0 +1,59 @@
...
+            name: 'ScanCompleted',
+            properties: eventProperties,
+        } as TelemetryEvent);
+    }
+
+    public async failRun(): Promise<void> {
+        // We don't send anything for failed runs
+    }
+
+    public didScanSucceed(): Promise<boolean> {
```
karanbirsingh, line 56: do we need this method if always `true`? https://github.com/microsoft/accessibility-insights-action/blob/6c9c25e5f92f495b9ca2969d6dbff126f0a5a50f/packages/shared/src/progress-reporter/progress-reporter.ts#L23

packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
brocktaylor7: This is a good catch, I've removed these properties.

packages/ado-extension/src/telemetry/app-insights-telemetry-client.ts
ThanyaLeif: I'm ok with getting rid of it, just asking for learning purposes.

packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
brocktaylor7: `reduce` is a built-in function in JavaScript that allows you to functionally compute a single value from an array. It's similar to `.map()` in that it functionally iterates over an array, but it does something very different than map. 

I've updated the variable values to make it clearer what is going on, but essentially the first variable in a reducer function is an aggregator. Meaning it keeps a running tally of whatever value you're trying to calculate.  The second variable is the current item being iterated over in the array. So in this case it's counting the failure instances. `a` keeps a running tally of failures already iterated over, `b` is the current failed rule. It then checks how many urls are part of the current failed rule, and adds that number to the current tally (`a`). Once it has iterated over everything it will return a value equal to the total number of urls (or failed rule instances) in the entire array of results. 

MDN does a pretty good job explaining it with their write-up: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

Let me know if you have any other questions about it!

packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
brocktaylor7: You're correct, I've removed it!

packages/ado-extension/src/progress-reporter/telemetry/telemetry-sender.ts
ThanyaLeif: Thanks for the detailed explanation!
I understand. :)

ThanyaLeif (approved):
LGTM!

PR #1135, opened 2022-04-13
feat(ado-ext-telemetry): update staging validation pipeline with v2 changes

dbjorge:
#### Details

Release 2.0.0 bumped the major version of the staging extension to 2.0.0, but our staging validation workflow is still running against version `@1`. This updates the validation template to match the updated number.

Several of the scenarios were impacted by breaking changes, and there were a few new scenarios that make sense for us to validate. This PR also updates the scenarios to be based on the new v2 functionality.

##### Motivation

Enable release validation for 2.0.0

##### Context

The syntax in the template was updated already in #1122

#### Pull request checklist

- [x] Addresses an existing issue: Story 1927731
- [n/a] Added relevant unit test for your changes. (`yarn test`)
- [n/a] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`yarn precheckin`)

PR #1141, opened 2022-04-19
feat(ado-ext-telemetry): update outputDir to write to a temp dir by default

dbjorge:
#### Details

This PR updates the ADO `outputDir` input to be optional without a default, instead of defaulting to `_accessibility-reports`, and to create a uniquely generated temp directory to use for `outputDir` if none is specified.

This PR intentionally does not update the GitHub Action with the same behavior because we didn't implement the embedded artifact uploading there, and it doesn't make sense to use a temp dir without the automatic artifact uploading. I did put the interesting part of the temp file generation in the `shared` package - I think if we wanted to implement it in the action in the future, it might makes sense to move the current `AdoTaskConfig.getReportOutDir` implementation into the base `TaskConfig` class as a wrapper around private-protected `getRawOutputDir` methods overriden by the subclasses.

##### Motivation

Now that uploading the output as an artifact is default, embedded behavior, the `outputDir` is an implementation detail for most users. We noticed in #1135 that it felt awkward to have to specify an implementation detail just to run the task multiple times from one pipeline, so this PR updates the default behavior to avoid having to do that.

This is technically breaking, since a user could previously have had a publish step that relied on the default _accessibility-reports directory, so it's been documented in the v1 to v2 docs.

##### Context

I think it would be nice for us to detect and give a nice error in the case where `outputDir` is specified explicitly in a way which is not unique among multiple steps in the same pipeline (probably by detecting if it already exists and erroring if so). I've left that out of scope for this PR since I expect it might make sense to fold it into the other input validation stuff @ThanyaLeif is in progress on.

#### Pull request checklist

- [x] Addresses an existing issue: 1940706

### DIFF
--- a/docs/ado-extension-usage.md
+++ b/docs/ado-extension-usage.md
@@ -197,7 +197,7 @@ Version 2.x of the extension contains several breaking changes from version 1.x.
     - If you previously specified just `url` and not `siteDir`, you should leave the original `url` input as-is
 3. Publishing a pipeline artifact containing scan results is now built into the Accessibility Insights task, instead of being a separate step you must add yourself afterwards
     - If you previously used a separate `publish` step to upload the `_accessibility-reports` folder, you can delete that `publish` step
-    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, specify `uploadOutputArtifact: false` to skip the new automatic artifact uploading and specify an `outputDir` to control where the output artifact contents get written to on the build agent
+    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, specify `uploadOutputArtifact: false` to skip the new automatic artifact uploading. You can specify `outputDir` to control where the output artifact contents get written to on the build agent
     - See [Report Artifacts](#report-artifacts) for more details, including how to customize the artifact name
 4. By default, the task now fails if it detects an accessibility failure (unless the failure is a known issue tracked by a [Baseline File](#using-a-baseline-file))
     - If you previously specified `failOnAccessibilityError: true`, you can remove it (this is now the default behavior)
@@ -216,7 +216,7 @@ Version 2.x of the extension contains several breaking changes from version 1.x.
     - If you previously specified _both_ as "Site Directory" and a "Website URL", you had a misconfiguration - these options were mutually exclusive, and the "Website URL" option was being silently ignored. Select `staticSite` mode and ignore your old "Website URL" input
 3. Publishing a pipeline artifact containing scan results is now built into the Accessibility Insights task, instead of being a separate step you must add yourself afterwards
     - If you previously used a separate "Publish" step to upload the `_accessibility-reports` folder, you can delete that "Publish" step
-    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, uncheck the "Upload Output Artifact" option to skip the new automatic artifact uploading and specify an "Output Directory" to control where the output artifact contents get written to on the build agent
+    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, uncheck the "Upload Output Artifact" option to skip the new automatic artifact uploading. You can specify an "Output Directory" to control where the output artifact contents get written to on the build agent
     - See [Report Artifacts](#report-artifacts) for more details, including how to customize the artifact name
 4. The "Fail on Accessibility Error" option is now checked by default; when it is checked, the task will fail if it detects an accessibility failure (unless the failure is a known issue tracked by a [Baseline File](#using-a-baseline-file))
     - If you would prefer to keep the old behavior, where accessibility issues are not treated as a task failure, you can still uncheck this option (but consider [using a Baseline File](#using-a-baseline-file) instead!)
--- a/docs/ado-extension-usage.md
+++ b/docs/ado-extension-usage.md
@@ -156,9 +156,11 @@ Here is an example of a YAML file that is configured to take advantage of a base
 
 ## Report Artifacts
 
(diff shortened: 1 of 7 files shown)


docs/ado-extension-usage.md
```diff
@@ -173,7 +175,7 @@ You can choose to block pull requests if the extension finds accessibility issue
 
 ## Running multiple times in a single pipeline
 
-If you want to run multiple Accessibility Insights steps in a single pipeline, you will need to ensure that each step uses a unique `outputArtifactName` and `outputDir`. Artifact names and output directories must be unique across all steps in a pipeline.
+If you want to run multiple Accessibility Insights steps in a single pipeline, you will need to ensure that each step uses a unique `outputArtifactName`. Any custom `outputDir` settings must also be unique among all steps in a pipeline.
```
karanbirsingh, line 178: When I first read this, I considered suggesting "...each step uses a unique `outputArtifactName` and `outputDir`". But then I remembered your phrasing is important because it indicates `outputDir` only needs to be unique if provided (the default is the temp directory). It's more verbose but might be nice to add an additional sentence like "...uses a unique `outputArtifactName`. The task will generate unique `outputDir` values by default, but if you override `outputDir`, you need to ensure they are unique among all steps in a pipeline."

docs/ado-extension-usage.md
```diff
@@ -212,7 +214,7 @@ Version 2.x of the extension contains several breaking changes from version 1.x.
     - If you previously specified _both_ as "Site Directory" and a "Website URL", you had a misconfiguration - these options were mutually exclusive, and the "Website URL" option was being silently ignored. Select `staticSite` mode and ignore your old "Website URL" input
 3. Publishing a pipeline artifact containing scan results is now built into the Accessibility Insights task, instead of being a separate step you must add yourself afterwards
     - If you previously used a separate "Publish" step to upload the `_accessibility-reports` folder, you can delete that "Publish" step
-    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, uncheck the "Upload Results as Artifact" option to skip the new automatic artifact uploading
+    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, uncheck the "Upload Output Artifact" option to skip the new automatic artifact uploading and specify an "Output Directory" to control where the output artifact contents get written to on the build agent
```
karanbirsingh, line 217: small suggestion to avoid parsing "and specify" as part of "to skip"

```suggestion
    - If your pipeline is running in OneBranch, or any other environment where individual tasks cannot publish artifacts directly, uncheck the "Upload Output Artifact" option to skip the new automatic artifact uploading. Also specify an "Output Directory" to control where the output artifact contents get written to on the build agent
```

packages/ado-extension/src/task-config/ado-task-config.ts
```diff
@@ -5,23 +5,32 @@ import * as adoTask from 'azure-pipelines-task-lib/task';
...
     constructor(
         @inject(iocTypes.Process) protected readonly processObj: typeof process,
+        @inject(TempDirCreator) private readonly tempDirCreator: TempDirCreator,
         private readonly adoTaskObj = adoTask,
         private readonly resolvePath: typeof resolve = resolve,
     ) {
         super(processObj);
     }
 
+    // memoizing this is important to avoid generating multiple temp directories in the defualt case
```
karanbirsingh, line 23: (small typo)
```suggestion
    // memoizing this is important to avoid generating multiple temp directories in the default case
```

packages/ado-extension/src/task-config/ado-task-config.ts
```diff
@@ -5,23 +5,32 @@ import * as adoTask from 'azure-pipelines-task-lib/task';
...
+    private memoizedReportOutDir: string | null = null;
     public getReportOutDir(): string {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        return this.getAbsolutePath(this.adoTaskObj.getInput('outputDir'))!;
+        if (this.memoizedReportOutDir == null) {
+            const customOutputDir = this.getAbsolutePath(this.adoTaskObj.getInput('outputDir'));
+
+            this.memoizedReportOutDir =
+                customOutputDir != null ? customOutputDir : this.tempDirCreator.createTempDirSync(this.getVariable('Agent.TempDirectory'));
```
karanbirsingh, line 30: (small) Would nullish coalescing work here too? It feels a little more readable to me but either is fine.
```suggestion
            this.memoizedReportOutDir =
                customOutputDir ?? this.tempDirCreator.createTempDirSync(this.getVariable('Agent.TempDirectory'));
```

karanbirsingh (approved):
Left a few small comments

docs/ado-extension-usage.md
dbjorge: Yes, I struggled with how to word this and rewrote it a few times. I like your suggestion better than what I came up with, will update!

dbjorge: Reworded your suggestion somewhat to avoid confusion between the task generating a unique output directory vs a unique `outputDir` input (it technically does the former, not the latter)

docs/ado-extension-usage.md
dbjorge: Fixed here and in the equivalent section for classic pipelines.

packages/ado-extension/src/task-config/ado-task-config.ts
dbjorge: Good call; I had avoided it intentionally, thinking that it's not compatible with node < 14, but now that you've suggested it I've remembered that this doesn't matter since we're transpiling it with node10-compatible output settings anyway. (this isn't the first usage in the codebase)

PR #1149, opened 2022-04-21
feat(ado-ext-telemetry): Updating input validation to run before scan starts

ThanyaLeif:
#### Details

This PR is part of the validations for the ADO Extension release.
This PR aims to move input validation to an earlier stage of the extension workflow.

Previously validation was done inside `workflow-enforcer.ts` but this was problematic as the scanner itself makes use of input variables in order to proceed.
After some discussion it was decided to move the validation to `scanner.ts` file a shared a place for GH Action and ADO extension, preventing the scanner from being executed.
Improvements achieved with the new `input-validator.ts` file include:

- Input errors are detected early during the action's runtime before any scanning begins, and stop the action from trying to do any other work once detected
- Emit a case-specific error message
- Not emit a stack trace
- Include in the error message specific information about exactly which input names are wrong, and exactly what modifications the user would have to make to the inputs to correct it
- Include in the error message and links that redirect to the relevant usage docs section.

##### Motivation

This is part of the validations for the ADO extension release

##### Context

Another option I considered was to create `InputValidator` as an abstract class that would be implemented on both sides (GH Action and ADO extension) in a similar way to `TaskConfig` in order to get the specific names of the variables to display in the error message. But I ended up adding a helper function `getInputName` to `task-config.ts` that translates the variable names to avoid overengineering.

#### Pull request checklist


packages/shared/src/scanner/scanner.ts
```diff
@@ -45,10 +46,14 @@ export class Scanner {
         @inject(BaselineOptionsBuilder) private readonly baselineOptionsBuilder: BaselineOptionsBuilder,
         @inject(BaselineFileUpdater) private readonly baselineFileUpdater: BaselineFileUpdater,
         @inject(iocTypes.TelemetryClient) private readonly telemetryClient: TelemetryClient,
+        @inject(InputValidator) private readonly inputValidator: InputValidator,
         private readonly fileSystemObj: typeof fs = fs,
     ) {}
 
     public async scan(): Promise<ScanSucceededWithNoRequiredUserAction> {
+        if (!(await this.inputValidator.validate())) {
+            return Promise.resolve(false);
```
dbjorge, line 55: `scan()` is an `async` function, so it already handles wrapping results in a `Promise` for you; you can just `return false;` here instead of `return Promise.resolve(false);`.

Letting the async machinery handle adding the `Promise` wrapper for you is a better general practice because if an `Error` gets thrown that involves the `Promise`, it will have a stack trace that's easier to read.

packages/ado-extension/scripts/local-overrides.json
```diff
@@ -1,4 +1,8 @@
 [
+    {
+        "name": "url",
+        "value": "https://www.washington.edu/accesscomputing/AU/before.html"
+    },
```
brocktaylor7, line 5: I think it makes sense to leave the `url` input out of this file and allow the `staticSiteDir` to handle the local runs by default. This configuration does successfully prove that these changes work but prevents the extension from scanning locally in a dev environment without changes to this file and should therefore probably be removed.

I also think changing the local default of `failOnAccessibilityError` to `true` below in this same file doesn't make the most sense for development purposes.

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+                this.logger.logError(errorMessage);
+                await this.failConfiguration();
+                return true;
+            }
+        }
+        await this.failConfiguration();
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
```
dbjorge, line 128: I think the eslint error being suppressed here is valid; it's explaining that there isn't any reason for this to be async. I think this is true of the whole file; I don't think anything in this class needs to be async.

It's preferable to avoid using `async` for functions that are entirely synchronous; it makes it easier for a reader of the code to reason about which operations are actually likely to be time consuming. eg, seeing an `await inputValidator.validate()` call in `scanner.scan()` outside of its timeout enforcement logic made me wonder "how long can input validation actually take in practice? Does it need to be inside a timeout?", when it actually isn't doing any long-running work.

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { injectable, inject } from 'inversify';
+import { iocTypes } from './ioc/ioc-types';
+import { Logger } from './logger/logger';
+import { TaskConfig } from './task-config';
+import { sectionSeparator, link } from './console-output/console-log-formatter';
+@injectable()
+export class InputValidator {
+    private configurationSucceeded = true;
```
dbjorge, line 10: I don't think it's necessary for this class to be stateful; right now, it's basically single-use, since this field is never reset anywhere, which is surprising to me as a reader.

I think it would be preferable to make `configurationSucceeded` a local variable of the `validate()` method, have the individual `failIf*` methods communicate results by returning rather than setting a field, and have `validate` invoke them using a pattern like `isValid &&= this.verifyAHostingModeCanBeInferred();`

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+    public async failConfiguration(): Promise<void> {
+        this.configurationSucceeded = false;
+    }
+
+    private writeConfigurationError(errorCase: string, errorInfo?: string): string {
+        const description = [
+            errorCase,
+            errorInfo,
+            'For more information visit:',
+            link('https://github.com/microsoft/accessibility-insights-action/blob/main/docs/gh-action-usage.md', 'GH Action documentation'),
```
dbjorge, line 138: I think it would be preferable to just link the relevant one of these. I'd do this by adding a `getUsageDocsUrl()` to `TaskConfig` and making the text of the link platform-agnostic (eg, `link(this.taskConfig.getUsageDocsUrl(), 'usage documentation')`)

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+                const failedInputNames = failedInputs.join(', ');
+                const errorCase = `A configuration error has ocurred ${failedInputNames} must not be set when dynamic mode is selected`;
+                const errorInfo = `To fix this error make sure ${failedInputNames} has not been set in the input section of your YAML file`;
+                const errorMessage = this.writeConfigurationError(errorCase, errorInfo);
+                this.logger.logError(errorMessage);
+                await this.failConfiguration();
+                return true;
+            }
+        }
+        await this.failConfiguration();
```
dbjorge, line 124: This will result in the input validator always failing if `hostingMode` is not dynamic, no? I think this probably suggests that there are some missing unit tests.

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public async failConfiguration(): Promise<void> {
+        this.configurationSucceeded = false;
+    }
+
+    private writeConfigurationError(errorCase: string, errorInfo?: string): string {
+        const description = [
+            errorCase,
+            errorInfo,
+            'For more information visit:',
```
dbjorge, line 137: It feels a bit off to repeat the usage message if there are multiple configuration errors. Maybe instead, the top level `validate()` method should print the usage message (just once) if any error(s) occurred?

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+        return Promise.resolve(this.configurationSucceeded);
+    }
+
+    private async failIfSiteDirAndUrlAreNotConfigured(): Promise<boolean> {
+        const url = this.taskConfig.getUrl();
+        const siteDir = this.taskConfig.getStaticSiteDir();
+        if (url === undefined && siteDir === undefined) {
+            const siteDirName = this.taskConfig.getInputName('StaticSiteDir');
+            const errorCase = `A configuration error has occurred url or ${siteDirName} must be set`;
+            const errorMessage = this.writeConfigurationError(errorCase);
```
dbjorge, line 32: This pattern feels a little redundant, it gets repeated in each of these helpers:

```ts
            const errorMessage = this.writeConfigurationError(errorCase);
            this.logger.logError(errorMessage);
            await this.failConfiguration();
            return true;
```

Maybe instead `writeConfigurationError` could be responsible for doing the `logError` (its name suggests that it should be, anyway), and combined with the other comment suggesting replacing the stateful `failConfiguration()` with just using return values, reduce this pattern to just:

```ts
            return this.indicateConfigurationError(errorCase);
```

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+        await this.failConfiguration();
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public async failConfiguration(): Promise<void> {
+        this.configurationSucceeded = false;
+    }
+
+    private writeConfigurationError(errorCase: string, errorInfo?: string): string {
```
dbjorge, line 133: I felt a little confused by the `errorCase` vs `errorInfo` naming; it wasn't clear to me what the difference was without reading the implementation of the function. Maybe more readable to just do `...errorLines: string[]`?

packages/shared/src/scanner/scanner.ts
ThanyaLeif: Addressed in last commit :)

packages/ado-extension/scripts/local-overrides.json
ThanyaLeif: Addressed in last commit :)

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return Promise.resolve(this.configurationSucceeded);
+    }
+
+    private async failIfSiteDirAndUrlAreNotConfigured(): Promise<boolean> {
+        const url = this.taskConfig.getUrl();
+        const siteDir = this.taskConfig.getStaticSiteDir();
+        if (url === undefined && siteDir === undefined) {
+            const siteDirName = this.taskConfig.getInputName('StaticSiteDir');
+            const errorCase = `A configuration error has occurred url or ${siteDirName} must be set`;
```
dbjorge, line 31: nit: it would be good to use `getInputName` for `url` as well, for consistency's sake

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return false;
+    }
+
+    private async failIfSiteDirAndUrlAreConfigured(): Promise<boolean> {
+        const url = this.taskConfig.getUrl();
+        const siteDir = this.taskConfig.getStaticSiteDir();
+        if (url !== undefined && siteDir !== undefined) {
+            const siteDirName = this.taskConfig.getInputName('StaticSiteDir');
+            const errorCase = `A configuration error has ocurred only one of the following inputs can be set at a time: url or ${siteDirName}`;
```
dbjorge, line 45: sp: `ocurred` -> `occurred`

packages/shared/src/input-validator.ts
dbjorge, line 31: It would be good to separate the `A configuration error has occurred` part and the individual message part with a colon; right now it reads like a run-on sentence.

packages/shared/src/input-validator.ts
ThanyaLeif: Addressed.

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+        return false;
+    }
+
+    private async failIfSiteDirIsNotConfiguredInStaticMode(): Promise<boolean> {
+        const siteDir = this.taskConfig.getStaticSiteDir();
+        const hostingMode = this.taskConfig.getHostingMode();
+        if (hostingMode === 'staticSite' && siteDir === undefined) {
+            const siteDirName = this.taskConfig.getInputName('StaticSiteDir');
+            const errorCase = `A configuration error has ocurred ${siteDirName} must be set when static mode is selected`;
+            const errorInfo = `To fix this error make sure to add ${siteDirName} to the input section in the corresponding YAML file`;
```
dbjorge, line 60: I think you can just omit the second line here. Some of our users are using ADO's "Classic" Pipelines, and don't have a YAML file.

packages/shared/src/input-validator.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return false;
+    }
+
+    private async failIfSiteDirIsNotConfiguredInStaticMode(): Promise<boolean> {
+        const siteDir = this.taskConfig.getStaticSiteDir();
+        const hostingMode = this.taskConfig.getHostingMode();
+        if (hostingMode === 'staticSite' && siteDir === undefined) {
+            const siteDirName = this.taskConfig.getInputName('StaticSiteDir');
+            const errorCase = `A configuration error has ocurred ${siteDirName} must be set when static mode is selected`;
```
dbjorge, line 59: A user might not be clear on what it means for "static mode" is referring to the hosting mode input; maybe "...when ${hostingModeName} is set to static"?

dbjorge: (here and in the similar messages in other helper functions)

packages/shared/src/input-validator.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    let taskConfigMock: IMock<TaskConfig>;
+    let loggerMock: IMock<Logger>;
+    let inputValidator: InputValidator;
+
+    beforeEach(() => {
+        taskConfigMock = Mock.ofType<TaskConfig>(undefined, MockBehavior.Strict);
+        loggerMock = Mock.ofType<Logger>(undefined, MockBehavior.Strict);
+    });
+
+    describe('constructor', () => {
```
karanbirsingh, line 20: I think it might be OK to forgo having tests specific to both `gh-action` usage and specific to `ado-extension` usage. We may get equal test coverage with just one - you could imagine generic testing code like `setupInputName('StaticSitePortInputName', 'StaticSitePort').

One reason separating the tests could seem helpful is that we can pass in `localhost-port` for the gh-action case and `staticSitePort` for the ado-extension case - but from the perspective of `input-validator`, these scenarios follow the same code path twice (just with different input values). `input-validator` only has a dependency on the `taskConfig` abstraction.

For example, we might come back in a future PR and decide to change GH-action `localhost-port`'s input name to `staticSitePort` so that we're consistent across both products. That code change would be inside `gh-task-config`, not `input-validator` - and it might feel odd then to change the unit tests for `input-validator`. Instead, we might want to update any existing unit tests for `gh-task-config.ts`.

packages/shared/src/input-validator.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+        });
+
+        it('initializes in gh scanner', () => {
+            inputValidator = buildInputValidatorWithMocks();
+            verifyAllMocks();
+        });
+    });
+
+    describe('validate gh-action', () => {
+        it('input configuration fail if siteDir and url are set at the same time', async () => {
```
dbjorge: It might be easier to write/reason about/update these test cases if you used a recording logger and parameterized tests with snapshots of the recorded error messages instead. Maybe something like:

```ts
it.each`
    hostingMode    | staticSiteDir   | url
    ${ undefined } | ${ 'site-dir' } | ${ 'url' }
`(
    'returns false with pinned error logs when hostingMode=$hostingMode, staticSiteDir=$staticSiteDir, url=$url',
    (taskConfigInputs) => {
        setupTaskConfigInputs(taskConfigInputs);
        expect(inputValidator.validate()).toBe(false);
        expect(recordingLogger.errorMessages).toMatchSnapshot();
    }
);

it.each`
    hostingMode    | staticSiteDir   | url
    ${ undefined } | ${ 'site-dir' } | ${ undefined }
`(
    'returns true with no error logs when hostingMode=$hostingMode, staticSiteDir=$staticSiteDir, url=$url',
    (taskConfigInputs) => {
        setupTaskConfigInputs(taskConfigInputs);
        expect(inputValidator.validate()).toBe(true);
        expect(recordingLogger.errorMessages).toStrictEqual([]);
    }
);
```

packages/shared/src/input-validator.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    let loggerMock: IMock<Logger>;
+    let inputValidator: InputValidator;
+
+    beforeEach(() => {
+        taskConfigMock = Mock.ofType<TaskConfig>(undefined, MockBehavior.Strict);
+        loggerMock = Mock.ofType<Logger>(undefined, MockBehavior.Strict);
+    });
+
+    describe('constructor', () => {
+        it('initializes in ado scanner', () => {
```
dbjorge, line 21: I'm not sure the different "ado" vs "gh" cases throughout these tests make sense anymore now that you've updated the class to move the implementation differences out into the different `TaskConfig` implementations

packages/shared/src/input-validator.spec.ts
dbjorge: (you can see an example of the "recording logger" pattern in the web repo at `/src/tests/unit/common/recording-logger.ts`)

packages/shared/src/scanner/scanner.spec.ts
```diff
@@ -27,6 +27,7 @@ import { CombinedReportParameters } from 'accessibility-insights-report';
 import { TaskConfig } from '../task-config';
 import * as fs from 'fs';
 import { TelemetryClient } from '../telemetry/telemetry-client';
+import { InputValidator } from '../input-validator';
 
 describe(Scanner, () => {
```
dbjorge, line 32: It would be good to add a test case here to verify that if the input validator returns false, the scanner doesn't try to do any actual work. That was the original issue motivating this PR, so it would be a particularly good test case to have.

packages/shared/src/ioc/setup-ioc-container.ts
```diff
@@ -30,6 +31,7 @@ export function setupSharedIocContainer(container = new inversify.Container({ au
     setupCliContainer(container);
 
     container.bind(Scanner).toSelf().inSingletonScope();
+    container.bind(InputValidator).toSelf().inSingletonScope();
```
dbjorge, line 34: I don't think this should be necessary; the input validator's behavior doesn't rely on it being a singleton and it isn't expensive enough to warrant making sure we avoid creating extras.

packages/shared/src/input-validator.spec.ts
dbjorge: Your version of the same comment was better, I removed mine :)

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.ts
ThanyaLeif: Addressed

packages/shared/src/input-validator.ts
ThanyaLeif: Addressed

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.ts
ThanyaLeif: Addressed

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.spec.ts
ThanyaLeif: Addressed.
It used to make sense to have constructor validation since my first approach was to send a parameter to detect where the scanner was called (ADO vs GH), but after refactoring it doesn't make sense.
I already remove one, thanks for the detailed explanation :)

packages/shared/src/scanner/scanner.spec.ts
ThanyaLeif: Addreesed

packages/shared/src/input-validator.spec.ts
ThanyaLeif: I'll address this comment in a future PR

packages/shared/src/input-validator.ts
ThanyaLeif: Addreesed

packages/shared/src/scanner/scanner.spec.ts
```diff
@@ -97,9 +101,16 @@ describe(Scanner, () => {
     });
 
     describe('scan', () => {
+        it('scanner is not initialized if input validation fails', async () => {
+            inputValidatorMock.setup((m) => m.validate()).returns(() => false);
+
+            const result = await scanner.scan();
+            expect(result).toBe(false);
```
dbjorge, line 108: This is testing that the scanner indicates failure, but not that the "scan" and "local file server" steps are skipped. A reasonable way to test that might be to add mock verifications like this to the end of the test:

```ts
aiCrawlerMock.verify(m => m.crawl(It.isAny(), It.isAny()), Times.never());
localFileServerMock.verify(m => m.start(), Times.never());
```

packages/shared/src/input-validator.spec.ts
dbjorge: It looks like this is only partially addressed; this comment applies to the duplicated `describe('validate gh-action', ...)` vs `describe('validate ado-extension', ...)` blocks, too, not just the constructor.

I think it would also be good to drop the "gh" part from the test name; it isn't meaningful anymore.

ThanyaLeif: My fault, removed in commit [c9ae5ee](https://github.com/microsoft/accessibility-insights-action/pull/1149/commits/c9ae5eefe619463b8243b559ea1d63d37e0ff81b)

packages/shared/src/scanner/scanner.spec.ts
ThanyaLeif: Addressed in commit [0f28c1f](https://github.com/microsoft/accessibility-insights-action/pull/1149/commits/0f28c1fa5a62d8804a1d281243f23b2e051d4e1e)

dbjorge (approved):

PR #1160, opened 2022-04-26
fix(ado-ext-telemetry): fix failOnAccessibilityError interaction with baseline files

dbjorge:
#### Details

The new `failOnAccessibilityError: true` default behavior interacts poorly with baseline files; it causes accessibility errors which are baselined to still fail the build, which defeats the purpose of the baseline file.

This PR updates the workflow enforcer behavior to ignore `failOnAccessibilityError` entirely if a baseline file is specified, instead.

It also improves the error output in the default case (no baseline file, default `failOnAccessibilityError` setting); previously, an accessibility error in that state would produce build log error output like this:

```
##[error]An accessibility error was found and you specified the "failOnAccessibilityError" flag.
```

This is misleading, because the common case is now going to be that the user actually didn't specify `failOnAccessibilityError` themselves at all. The error now instead reads:

```
##[error]Accessibility error(s) were found
To prevent accessibility errors from failing your build, you can:
* Use a baseline file to avoid failing for known issues, or
* Set the failOnAccessibilityError task input to false to avoid failing for all issues
```

I also introduced a new test utility class (`RecordingTestLogger`) to enable the use of snapshots with the workflow enforcer error messages; this is something @ThanyaLeif and I considered adding in #1149 , so it seems like a repeatedly useful concept.

This is an ADO-only change; the workflow enforcer, baseline functionality, and `failOnAccessibilityError` inputs are all ADO-specific.

##### Motivation

Fix issues caught during feature release validation

##### Context

n/a

#### Pull request checklist

- [x] Addresses an existing issue: 1893588

### DIFF
--- a/packages/ado-extension/src/progress-reporter/enforcement/__snapshots__/workflow-enforcer.spec.ts.snap
+++ b/packages/ado-extension/src/progress-reporter/enforcement/__snapshots__/workflow-enforcer.spec.ts.snap
@@ -1,0 +1,22 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`WorkflowEnforcer completeRun with baseline and failOnAccessibilityError=false fails with pinned error log if baseline needs to be updated 1`] = `
+Array [
+  "[info] ##vso[task.logissue type=error;sourcepath=/some/file] The baseline file does not match scan results.",
+]
+`;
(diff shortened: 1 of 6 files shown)


ThanyaLeif (approved):
Nicely done 👍🏼

PR #1161, opened 2022-04-27
feat(ado-ext-telemetry): improve Extensions tab summary header

dbjorge:
#### Details

The new "job summary in the ADO Extensions tab" feature uses a section header based on the filename of the markdown file we upload as the summary. Before this PR, we always used `results.md` as the filename, so we always got a section header of `results`, which didn't look quite right.

This PR updates the filename to use the following format:

* `Accessibility Insights scan summary` if the `uploadOutputArtifact` is false
* `Accessibility Insights scan summary` if we are uploading to an artifact with the default name ("accessibility-reports")
* `Accessibility Insights scan summary (artifact-name)` if we are uploading to an artifact with a non-default name (either because a user specified `outputArtifactName` themselves or because this is a rerun of a job and we've used something like `accessibility-reports-2` because of that)

ADO only change, no impact on GHA

##### Motivation

Avoid the off-looking `results` header in the new summary UI

##### Context

n/a

#### Pull request checklist

- [x] Addresses an existing issue: 1893588


ThanyaLeif (approved):
LGTM

PR #1163, opened 2022-04-28
fix(gha): prevent input validator treating unspecified url input as specified

dbjorge:
#### Details

In main right now, the GitHub action self test workflow (which specifies a `site-dir` and no `url`) fails with a message like this:

```
A configuration error has occurred, only one of the following inputs can be set at a time: url or site-dir
```

This happens because the input validator assumed (based on `TaskConfig` typings) that an unspecified url input would be exposed from `TaskConfig` as `undefined`. However, GitHub Actions' `getInput` return empty string rather than undefined for unspecified inputs, and `GHTaskConfig` is currently inconsistent about whether it normalizes those as `undefined` or not.

This PR updates `GHTaskConfig` to normalize unspecified parameters as `undefined` consistently. It also updates a few stray `TaskConfig` return types that should have indicated that they are optional, but didn't.

Right now, these typings are just informative, not enforced. Ideally, we'd enable strict null checks to enforce these at compile time.

No ADO impact.

##### Motivation

Fix blocking GitHub action bug

##### Context

n/a

#### Pull request checklist


dbjorge:
Reopening PR to attempt to fix CLA bot

brocktaylor7 (approved):
LGTM!

PR #1181, opened 2022-05-09
fix: staticSiteUrlRelativePath validation issue

brocktaylor7:
#### Details

This fixes an issue with input validation that was causing the following error when `staticSiteUrlRelativePath` is not set in `dynamicSite` hosting mode:

`##[error]A configuration error has occurred, staticSiteUrlRelativePath must not be set when hostingMode is set to dynamic`

This is due to the fact that in `task.json` we have a `defaultValue` set for that input. The validator picks up the default value and assumes that the user has supplied an input.

To fix this, we have allowed the default value of `/` to be an acceptable input in the input validator. That way if there is an invalid input that we know was added by the end user it will still throw the error. This leaves the existing default value in place so as not to break existing static site setups that do not specify the input and consume the default value.

This impact should be the same for both the ADO extension and the Github action, as they both pick up the relevant variable and default value.

##### Motivation

fix validation issues

##### Context

We considered removing the validation for that input altogether, but decided to simply allow for the default of `/` to be accepted by the validator. 

#### Pull request checklist

### DIFF
--- a/packages/shared/src/input-validator.spec.ts
+++ b/packages/shared/src/input-validator.spec.ts
@@ -109,7 +109,7 @@ describe(InputValidator, () => {
         it('input configuration fail if static inputs are set in dynamicSite mode', () => {
             setUpHostingMode('dynamicSite');
             setupGetStaticSiteDir('site-dir');
-            setupGetStaticSiteUrlRelativePath(defaultStaticSiteUrlRelativePath);
+            setupGetStaticSiteUrlRelativePath('non-default-static-site-url-relative-path');
             setupGetStaticSitePort(100);
             setupGetUrl('url');
             setupInputName('staticSiteDir', 'StaticSiteDir');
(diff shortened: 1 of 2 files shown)


ThanyaLeif (approved):
LGTM!

packages/shared/src/input-validator.spec.ts
```diff
@@ -108,15 +109,15 @@ describe(InputValidator, () => {
         it('input configuration fail if static inputs are set in dynamicSite mode', () => {
             setUpHostingMode('dynamicSite');
             setupGetStaticSiteDir('site-dir');
-            setupGetStaticSiteUrlRelativePath('ure-relative-path');
+            setupGetStaticSiteUrlRelativePath(defaultStaticSiteUrlRelativePath);
```
dbjorge, line 112: I think this one should still be using a non-default value, no?

brocktaylor7: Good call, I've added this back in!

PR #1204, opened 2022-06-10
feat(ado-gha-axe-core): Add Job Summary to GH Action

brocktaylor7:
#### Details

This PR adds the new Job Summary functionality to the Github Action

Here is a link to a test action run that uses the job summary, you can find it under "build summary" at the bottom of the summary tab:
https://github.com/brocktaylor7/accessibility-insights-action-test/actions/runs/2471620857

image

##### Motivation

Feature work.

##### Context

I spent some time looking into whether we could directly link the artifact within the job summary. Because the artifact upload happens in a subsequent step in the Github Action, the artifact doesn't yet exist when the job summary is being created and thus can't be linked.

There will be a separate PR that removes the check run and token. We may also need to look into whether we need to fail the build in that step when accessibility issues are found, as I believe the current place that the build fails is within the check run step.

This PR has no impact on the ADO Extension.

#### Pull request checklist

### DIFF
--- a/packages/gh-action/src/task-config/gh-task-config.spec.ts
+++ b/packages/gh-action/src/task-config/gh-task-config.spec.ts
@@ -12,14 +12,14 @@ describe(GHTaskConfig, () => {
     let processStub: any;
     let actionCoreMock: IMock<typeof actionCore>;
     let taskConfig: GHTaskConfig;
-    let markdownSummaryMock: IMock<typeof actionCore.markdownSummary>;
+    let markdownSummaryMock: IMock<typeof actionCore.summary>;
 
     beforeEach(() => {
         processStub = {
             env: {},
         } as any;
         actionCoreMock = Mock.ofType<typeof actionCore>();
-        markdownSummaryMock = Mock.ofType<typeof actionCore.markdownSummary>();
+        markdownSummaryMock = Mock.ofType<typeof actionCore.summary>();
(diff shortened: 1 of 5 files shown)


packages/gh-action/src/task-config/gh-task-config.spec.ts
```diff
@@ -12,17 +12,20 @@ describe(GHTaskConfig, () => {
     let processStub: any;
     let actionCoreMock: IMock<typeof actionCore>;
     let taskConfig: GHTaskConfig;
+    let markdownSummaryMock: IMock<typeof actionCore.markdownSummary>;
```
lisli1, line 15: It looks like actionCore.markdownSummary is deprecated (https://github.com/actions/toolkit/pull/1073). Should we explore using the suggested core.summary in place of this?

lisli1: Digging a little deeper into `Summary`, it looks like both point to the same object so we can just rename this to `summary` instead of `markdownSummary`.

lisli1 (approved):
Minor comment but LGTM!

packages/gh-action/src/task-config/gh-task-config.spec.ts
brocktaylor7: Great catch! I've updated this now :)

PR #1213, opened 2022-06-22
docs: add information about test deployments to GitHub actions

karanbirsingh:
#### Details

There are a few different ways to test updates to the GitHub action. One way is to deploy the built action remotely. I am adding some information to the docs so that it's in one place. It would be nice to operationalize this into our yarn scripts, but I'm not sure I'll be able to get to that immediately - so hoping that this provides some initial info.

As an example, running the listed commands produces this repo: https://github.com/karanbirsingh/accessibility-insights-action-test-deployment which has its own action.yml. So there is no second 'client' repository required - when the repo is pushed, it immediately begins self-testing its action.

The script is inside a Markdown file which can feel like an odd choice - I didn't put it into a PowerShell script. This is because of a few reasons:
- I am on a Windows machine and have not tested whether these work appropriately in other terminal/OS environments
- Since the commands are not encapsulated with error or retry logic, it didn't feel as 'ready' to put into a yarn command (especially with creating/deleting repos)
- I can easily imagine someone customizing which commands make sense for their workflow

In the "default case" you can copy-paste the full list into PowerShell after a `yarn build` and it results in a repo like the one I shared above.

##### Motivation

Share an approach that can be used to deploy and test the GH action remotely

#### Pull request checklist

- [ ] Described how this PR impacts both the ADO extension and the GitHub action: This PR only impacts the GH action.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -29,7 +29,7 @@ To make a change, you can follow these steps:
 
 # Deploy GitHub Action to your own test repo
 
-You can follow the instructions [here](../packages/gh-action/deploy-scripts/deploy-to-github-test.md).
+You can follow the [instructions to deploy to GitHub](../packages/gh-action/deploy-scripts/deploy-to-github-test.md).
 
 # Run GitHub Actions locally
 
--- a/dev/README.md
+++ b/dev/README.md
@@ -27,6 +27,10 @@ To make a change, you can follow these steps:
 -   push your changes to GitHub
 -   create a pull request. If your branch is on the main repo, the PR build should run your implementation against the test files in `website-root`.
 
+# Deploy GitHub Action to your own test repo
+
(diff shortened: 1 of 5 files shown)


dev/README.md
```diff
@@ -27,6 +27,10 @@ To make a change, you can follow these steps:
 -   push your changes to GitHub
 -   create a pull request. If your branch is on the main repo, the PR build should run your implementation against the test files in `website-root`.
 
+# Deploy GitHub Action to your own test repo
+
+You can follow the instructions [here](../packages/gh-action/deploy-scripts/deploy-to-github-test.md).
```
katydecorah, line 32: Small suggestion: descriptive link text can improve findability

```suggestion
You can follow the [instructions to deploy to GitHub](../packages/gh-action/deploy-scripts/deploy-to-github-test.md).
```

PR #1221, opened 2022-06-28
feat(ado-gha-axe-core): Remove repo token input

brocktaylor7:
#### Details

This removes the repo token input that was previously used to pull in the GH Octokit package allowing us to create/modify check runs. The token is still created in GitHub by default, and there is no way to disable this, as detailed here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication 
Removing the input makes it so our code doesn't have any access to it and removes any attack surface that may have existed by pulling it in as an input.

Here's a test run using my test action after these changes:
https://github.com/brocktaylor7/accessibility-insights-action-test/runs/7098823230?check_suite_focus=true

##### Motivation

Feature work

##### Context

ADO Extension impact: None, however we've now removed the tokens in both the GH Action and ADO Extension, so I've removed the `getToken` abstract method from the `TaskConfig` class.

#### Pull request checklist


JGibson2019 (approved):
LGTM

lisli1 (approved):

PR #1229, opened 2022-07-01
chore: remove unneeded dependency resolutions

pownkel:
#### Details

Remove dependency resolutions that are no longer needed

##### Motivation

Clean up package.json

#### Pull request checklist

- [ ] Described how this PR impacts both the ADO extension and the GitHub action

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,20 +68,11 @@
         "webpack-cli": "^4.10.0"
     },
     "resolutions": {
-        "apify-shared": ">=0.5.0",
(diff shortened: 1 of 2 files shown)


DaveTryon (approved):
I compared the versions of removed resolutions before and after the change. The only difference is that axios version 0.27.2 is used before the change, while 0.21.4 is used after the change. They're both high enough versions to satisfy known security issues, but I'm not sure if changing the version has any unintended consequences.

PR #1231, opened 2022-07-05
Fix: Adding hosting-mode to valid input list

ThanyaLeif:
#### Details

This PR adds `hosting-mode` to the valid list of inputs for GH Action and removes the default value for `static-site-dir`. It also changes `static-site-dir` and `static-site-url-relative-path` to be optional inputs, as they are not required in `dynamicSite` hosting mode.

Additionally updated the `gh-action-usage.md` to make the examples/suggestions more accurately reflect the new behavior and include a migration guide from v2 to v3, which specifies how to use `hosting-mode`.

##### Motivation
During the last release validation, @lisli1 found an issue where `hosting-mode` was not recognized as a valid input, preventing GH Action from performing an accessibility test on the project.

With the default `static-site-dir`, the GH Action wasn't able to run in `dynamicSite` hosting mode, since it expects `static-site-dir` to not be set when using `dynamicSite`. By removing the default, `static-site-dir` no longer has a value set. This is the same behavior as the ADO extension, which also does not have a default value for `staticSiteDir`.

##### Context
`hosting-mode` was not added to the list of valid inputs on the Action pipeline, causing configuration problems when set.

Considered keeping the existing behavior for `static-site-dir` with the default input and checking for it in `input-validator.ts`, but that would have required adding another method, `getSiteDirDefault` to the shared `task-config.ts`, which may also impact the ADO extension. Decided to keep the changes scoped to the GH Action and make the behavior match what is already in the ADO extension.

This PR exclusively affects GH Action, ADO Extension is not impacted.

#### Pull request checklist

### DIFF
--- a/docs/gh-action-usage.md
+++ b/docs/gh-action-usage.md
@@ -7,6 +7,7 @@ Licensed under the MIT License.
 
 To use this action in your workflow (which, again, we don't yet recommend at all for any production projects), we recommend referring to a version tag:
 
+-   To migrate from version 2, see [Migrating from version 2 to version 3](#migrating-from-version-2-to-version-3)
 -   `microsoft/accessibility-insights-action@v3` is updated with each `v3.x.y` release to refer to the most recent API-compatible version.
 -   `microsoft/accessibility-insights-action@v3.0.0` refers to the exact version `v3.0.0`; use this to pin to a specific version.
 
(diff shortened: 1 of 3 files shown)


JGibson2019 (approved):
Tested the new hosting-mode dynamicSite and it is working!

PR #1242, opened 2022-07-21
Update ado-extension-usage.md

asksep:
#### Details

"Modify crawling options" docs update to clarify behavior of 'maxUrls' input

##### Motivation

Wanted to add clarity around scanning a set number of URLs for the extension's crawl behavior

##### Context

Prior to the update, we had noted that if you set 'maxUrls' to 1, it would disable the crawler. This was a bit misleading, so we have updated the language. 

#### Pull request checklist

- [ ] n/a

### DIFF
--- a/docs/ado-extension-usage.md
+++ b/docs/ado-extension-usage.md
@@ -99,7 +99,7 @@ The action supports several crawling options defined in [task.json](https://gith
 For instance, you can:
 
 -   use `maxUrls: 1` to exclusively set the extension to only scan the first Url that has been inputted
--   set `maxUrls: # of inputUrls` to scan a set list of URLs for your site
+-   set `maxUrls` to the number of URLs in `inputUrls` to scan a fixed list of URLs for your site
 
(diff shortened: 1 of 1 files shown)


docs/ado-extension-usage.md
```diff
@@ -98,8 +98,8 @@ The action supports several crawling options defined in [task.json](https://gith
 
 For instance, you can:
 
--   use `maxUrls: 1` to turn off crawling
--   include a list of additional URLs to scan (the crawler won't find pages that are unlinked from the base page)
+-   use `maxUrls: 1` to exclusively set the extension to only scan the first Url that has been inputted
+-   set `maxUrls: # of inputUrls` to scan a set list of URLs for your site
```
dbjorge, line 102: Putting `# of inputUrls` in a code block like that might suggest to some readers that it's intended to be typed literally into their YAML config. Maybe more clear to instead do something like:

```suggestion
-   set `maxUrls` to the number of URLs in `inputUrls` to scan a fixed list of URLs for your site
```

dbjorge (approved):
Approved with 1 suggestion inline!

madalynrose (approved):

Issue #866, opened 2021-09-23
[ADO Extension] E2E tests


lisli1:
E2E tests were disabled in #844 due to an [issue in webpack](https://github.com/microsoft/azure-pipelines-task-lib/issues/794). When it is resolved, the tests will need to be turned back on.

asksep:
The GitHub Repo token usage is obsolete, therefor the Accessibility Insights team is going to investigate enabling new E2E tests

katydecorah:
I'm moving this backed to blocked. https://github.com/microsoft/azure-pipelines-task-lib updated `azure-pipelines-task-lib/mock-test` with Node 16 but hasn't release yet. This means that the e2e tests run on Node 10 and then fails when it runs the scan package (which requires Node 16).

PR #1262, opened 2022-08-10
feat(ado-auth-part-2): Upgrade ADO Extension to use Node16

brocktaylor7:
#### Details

This makes the necessary changes to upgrade our extension to use Node16. We reduced artifact uploads to just the report and baseline if applicable due to an issue with the Node16 handler on ADO.

I followed the guide for upgrading that the build task team pointed me to, found here: https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/migrateNode16.md

##### Motivation

Feature work

##### Context

The `azure-pipelines-task-lib` library still has the `-preview` tag. The build task team has done this intentionally to limit the number of pipelines that run on Node16 until they're able to scale up the number of available agents on Node16. The library is stable.

Several approaches were attempted with the artifact upload. I landed on manually uploading the report via logging commands, and checking for the baseline and uploading it in the same way if available. I've tested this in all three available OS environments (Windows, Mac, Ubuntu) and it is working as expected in all three.

This has no impact on the GH Action.

#### Pull request checklist

- [x] Ran prechecking (`yarn precheckin`)

### DIFF
--- a/packages/ado-extension/task.json
+++ b/packages/ado-extension/task.json
@@ -191,8 +191,9 @@
         }
     ],
     "execution": {
-        "Node10": {
+        "Node16": {
             "target": "index.js"
         }
-    }
+    },
+    "minimumAgentVersion": "2.206.1"
 }
--- a/packages/ado-extension/package.json
+++ b/packages/ado-extension/package.json
@@ -34,7 +34,7 @@
     "dependencies": {
         "@accessibility-insights-action/shared": "^1.0.0",
         "applicationinsights": "^2.3.4",
-        "azure-pipelines-task-lib": "^3.3.1",
+        "azure-pipelines-task-lib": "^4.0.0-preview",
         "reflect-metadata": "^0.1.13"
     }
 }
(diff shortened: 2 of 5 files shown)


packages/ado-extension/task.json
```diff
@@ -191,8 +191,9 @@
         }
     ],
     "execution": {
-        "Node10": {
+        "Node16": {
```
katydecorah, line 194: Sharing some small, non-blocking notes:
- Until the schema for v4 is released (https://github.com/microsoft/azure-pipelines-task-lib/pull/852), contributors may see a squiggle line with the tooltip `Property Node16 is not allowed` in their IDE for this value but it doesn't affect the functionality of the env.

- [That pull request](https://github.com/microsoft/azure-pipelines-task-lib/pull/852) updated the Node version for `azure-pipelines-task-lib/mock-test` which affects only our e2e test which is currently disabled (https://github.com/microsoft/accessibility-insights-action/issues/866). We'll need to wait for another task lib release to be able to add back the e2e tests after this merges.

katydecorah (approved):
Looks good to me!

Issue #1000, opened 2022-01-03
Jest configuration doesn't support pure-ESM packages


karanbirsingh:
We've recently seen some dependabot updates to packages that are pure-ESM. This issue lists updates whose builds failed & whose update path wasn't immediately obvious - I didn't want to lose track of them. Looking closer & updating this description as we go
- https://github.com/microsoft/accessibility-insights-action/pull/890
- https://github.com/microsoft/accessibility-insights-action/pull/970

At first I thought this was because this repo targets Common JS; in general, trying to import ESM packages into CommonJS requires [dynamic await() imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports). I thought we'd need to follow the steps listed here: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c 

After trying the update locally, I noticed webpack overrides the require function so that overall index.js is still functional. The PR build failures come from jest tests with the error signature: 

```
@accessibility-insights-action/shared: FAIL @accessibility-insights-action/shared src/ioc/setup-ioc-container.spec.ts
@accessibility-insights-action/shared:   ● Test suite failed to run
@accessibility-insights-action/shared:     Jest encountered an unexpected token
@accessibility-insights-action/shared:     Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
@accessibility-insights-action/shared:     Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.
@accessibility-insights-action/shared:     By default "node_modules" folder is ignored by transformers.
@accessibility-insights-action/shared:     Here's what you can do:
@accessibility-insights-action/shared:      • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
@accessibility-insights-action/shared:      • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
@accessibility-insights-action/shared:      • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
@accessibility-insights-action/shared:      • If you need a custom transformation specify a "transform" option in your config.
@accessibility-insights-action/shared:      • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
[...]

PR #1268, opened 2022-08-15
chore: Enable allowJs to support pure ESM with Jest

katydecorah:
#### Details

The shared package has pure ESM dependencies (get-port, serialize-error) which are not out-of-the-box compatible with Jest. This pull request: 

- enables `allowJs` in the tsconfig which will allow ts-jest to transform these modules. 
- sets `include` to the shared tsconfig to prevent extraneous js files from changing the compile file directory.
- removes the dependabot version pins for the dependencies and update them to their latest versions.

##### Motivation

Addresses #1000

##### Context

An alternative approach is to use babel-jest to transform the ESM dependencies. This would require two new dependencies to maintain and making sure regenerator-runtime runs before Jest starts up. One benefit with this solution is that it doesn't modify the tsconfig. [Example branch.](https://github.com/microsoft/accessibility-insights-action/compare/main...katydecorah:accessibility-insights-action:jest-esm-babel?expand=1)

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #0000

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -45,11 +45,3 @@ updates:
   - dependency-name: marked-terminal
     versions:
     - ">= 5"
-    # get-port v6+ is pure ESM and requires node >= 12.20
-  - dependency-name: get-port
-    versions:
-    - ">= 6"
-    # serialize-error v9+ is pure ESM and requires node >= 12.20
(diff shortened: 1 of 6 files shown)


dbjorge (approved):
LGTM!

PR #1283, opened 2022-08-30
chore: update ubuntu version to 20.04

katydecorah:
#### Details

Update the ubuntu version to latest to move off 18.04. I noticed the following warning:

> ##[warning]The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

##### Motivation

Keep pipeline environments up to date.

#### Pull request checklist

### DIFF
--- a/pipelines/build.yaml
+++ b/pipelines/build.yaml
@@ -7,7 +7,7 @@ trigger: # trigger only affects CI builds
             - main
 
 variables:
-    linuxImage: 'ubuntu-18.04'
+    linuxImage: 'ubuntu-20.04'
     windowsImage: 'windows-2019'
 
 jobs:
--- a/pipelines/build.yaml
+++ b/pipelines/build.yaml
@@ -7,7 +7,7 @@ trigger: # trigger only affects CI builds
             - main
 
 variables:
-    linuxImage: 'ubuntu-20.04'
+    linuxImage: 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


pipelines/build.yaml
```diff
@@ -7,7 +7,7 @@ trigger: # trigger only affects CI builds
             - main
 
 variables:
-    linuxImage: 'ubuntu-18.04'
+    linuxImage: 'ubuntu-20.04'
```
katydecorah, line 10: All other workflows use `ubuntu-latest`. Since this is the build task, I can see advantages to pinning the version rather than latest. However, setting `ubuntu-latest` would prevent continued maintenance of the version number here. Any strong opinions here?

ThanyaLeif (approved):

pipelines/build.yaml
brocktaylor7: I don't have strong opinions on it but I think I lean towards setting it to `ubuntu-latest`. If it breaks against newer versions of Ubuntu that's something that would be good for us to know as soon as possible. So it both prevents continued maintenance of version numbers and also could potentially alert us of problems early.

katydecorah, line 10:
```suggestion
    linuxImage: 'ubuntu-latest'
```

PR #1359, opened 2022-10-26
fix: Add empty option for authType

### LINKED ISSUE #1357: Cannot add task in a classic pipeline without setting Authentication type
## Describe the bug

A user reported that it's not possible to add the AI task to a classic pipeline without setting the Authentication Type. If you do not need authentication and are required to select the Authentication Type, then the task will fail with the error "--serviceAccountName, --serviceAccountPassword, and --authType all must be provided to scan authenticated pages."

**To Reproduce**

1. Go to ADO
2. Click "New Pipeline"
3. Click "Use the classic editor"
4. Click "Continue"
5. Click "Empty job"
6. Click the "+" On Agent Job 1
7. Add the AI Task
8. Click on the AI task and note that Authentication Type has an error "This setting is invalid. Select a valid option from the dropdown." and it's not possible to complete the task without selecting "Azure Active Directory" from the pipeline. The screenshot below illustrates this error. The dropdown has no other options.

## CodePen repro example

n/a

## Expected behavior

The user can submit an empty value for Authentication type if they do not require auth.

## Screenshots

Included in reproduction steps.

Classic ADO pipelines

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

katydecorah:
#### Details

Adds an empty option for authType in the ADO extension's task.json. This will fix an issue where classic pipeline users saw an invalid option error for the field.

This change does not impact the GitHub action.

##### Motivation

Address issue #1357 

##### Context

I pushed this branch to a staging task and was able to verify the expected behavior.

In the screenshot below. I am able to bypass the Authentication type error as the dropdown automatically selects the empty value. The YAML shows that `authType` is not set. I am able to save and queue the pipeline:

In the screenshot below, I am able to add an auth type and (fake) service account name and password. The YAML shows that `authType`, `serviceAccountName`, and `serviceAccountPassword` are set as expected.  I am able to save and queue the pipeline:

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1357 

### DIFF
--- a/packages/ado-extension/task.json
+++ b/packages/ado-extension/task.json
@@ -91,7 +91,7 @@
             "required": false,
             "defaultValue": "",
             "options": {
-                "": "",
+                "": "None",
                 "AAD": "Azure Active Directory"
             },
             "helpMarkDown": "Use with serviceAccountName and serviceAccountPassword to specify the authentication type."
--- a/packages/ado-extension/task.json
+++ b/packages/ado-extension/task.json
@@ -89,7 +89,9 @@
             "type": "pickList",
             "label": "Authentication type",
             "required": false,
+            "defaultValue": "",
             "options": {
+                "": "",
                 "AAD": "Azure Active Directory"
             },
             "helpMarkDown": "Use with serviceAccountName and serviceAccountPassword to specify the authentication type."


packages/ado-extension/task.json
```diff
@@ -89,7 +89,9 @@
             "type": "pickList",
             "label": "Authentication type",
             "required": false,
+            "defaultValue": "",
             "options": {
+                "": "",
```
dbjorge, line 94: Does this actually render in practice as a separate, selectable, blank dropdown option? That feels like it might be a little confusing (especially for a screen reader user) - maybe preferable to use a more explicit `"None"` option?

dbjorge (changes requested):
Thanks Katy! Comment inline

packages/ado-extension/task.json
katydecorah: Thanks for calling that in! I agree that using "None" as the label for the empty value is the right move.

katydecorah:
## Updated screenshots

In the screenshot below, the pipeline has set "Authentication type" to "None." The YAML does not set the authentication type. I am able to save and queue the build without using authentication:



In the screenshot below, the pipeline has set the service account name, password, and authentication type. The YAML shows the authentication inputs populated with the values (fake-account-name, fake-password, and AAD). I am able to save and queue the build while using authentication:

dbjorge (approved):

PR #1381, opened 2022-11-03
chore: add release change log generation tool

brocktaylor7:
#### Details

This adds a script to generate a changelog during release. This copies from the optional script we have for Accessibility Insights for Web releases.

It auto-generates the tester, verified, dev, commit, pr, change description, group, version, and date columns for all the changes between the current release and the previous.

To run it, in the root of the action repo you run the following command, for this example I'm using the previous two release tags:
`node ./tools/get-change-log-for-release.js --from v3.1.0 --to v3.1.1 `

This is the resulting file from that command:
[change-log.v3.1.0-v3.1.1.csv](https://github.com/microsoft/accessibility-insights-action/files/9933118/change-log.v3.1.0-v3.1.1.csv)

Once the table is generated and filtered, it looks like this (an example of what the previous release's table would have looked like):
image

##### Motivation

Additional tooling to simplify the release process

##### Context

This will work the same for both the GH action and the ADO Extension.

#### Pull request checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
         "npm-run-all": "^4.1.5",
         "prettier": "^2.7.1",
         "rimraf": "^3.0.2",
+        "simple-git": "^3.14.1",
         "ts-jest": "^29.0.3",
         "ts-loader": "^9.4.1",
         "typemoq": "^2.1.0",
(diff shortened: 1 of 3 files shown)


ThanyaLeif (approved):
LGTM!

PR #1384, opened 2022-11-04
fix: add process.exit to fix process hanging at the end of execution

brocktaylor7:
#### Details

This adds a `process.exit()` at the end of execution to ensure the task doesn't hang.

##### Motivation

Bug fix in preparation for release

##### Context

This is a bug that we had discovered during release validation that was causing the process to hang.  We discovered that the `accessibility-insights-scan` package was having the same thing happening so we fixed it and released that package upstream.  Perplexingly, the issue wasn't resolved by the upstream fix.

I have implemented a similar fix here by adding a `process.exit()` at the end of execution. I had initially considered adding it to a `finally` block within `packages/ado-extension/src/ado-extension.ts`, but it felt wrong to end the process one level down from the top-level call in `index.ts`. 

When I initially put the `process.exit()` in `index.ts`, it was exiting the process before the full inner async function in `ado-extension.ts` was complete, so I made the outer `runScan` function async, which allowed it to be awaited in `index.ts`, fixing the race condition.

I released my test extension with these changes and ran to verify that it works as expected: [Test run from my test extension](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=39047&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247)

This particular change only affects the ADO Extension. Later, we will need to test whether the issue also surfaces in the GH action.

#### Pull request checklist

### DIFF
--- a/packages/ado-extension/src/ado-extension-test.spec.ts
+++ b/packages/ado-extension/src/ado-extension-test.spec.ts
@@ -18,7 +18,7 @@ describe('runScan', () => {
         scanResponse = Promise.resolve(false);
         jest.spyOn(Logger.prototype, 'getAllErrors').mockReturnValueOnce(errorMessage);
         const setResultMock = jest.spyOn(adoTask, 'setResult');
-        runScan();
+        await runScan();
         await flushPromises();
         expect(setResultMock).toBeCalledWith(adoTask.TaskResult.Failed, errorMessage);
     });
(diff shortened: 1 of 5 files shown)


ThanyaLeif (approved):
The local run worked on my machine.
Everything looks normal... approved.

PR #1395, opened 2022-11-14
feat(docs-tests-telemetry): New Telemetry Error Sender

ThanyaLeif:
#### Details

This PR adds the new class `TelemetryErrorSender`, responsible for gathering and bundling errors from `InputValidator` and `Scanner`. This new class includes two main properties.
* `sender`: The class where the error was found, this could help us understand where the error is coming from.
* `errorList`: The list of error messages found in the class. Here we could find the same exact error message that was displayed to the user.

Here is an example of an `ErrorFound` event in our dev telemetry subscription, showing the `sender` and `errorList` properties mentioned above: 

image

##### Motivation

Store key information about users' errors while using our extension to help us find common issues and usage patterns.

##### Context

This new telemetry entry point will help us to develop a strategy to surface the most common issues faced by our users in future features.

The changes will be reflected exclusively on the ADO Extension side, with no user impact.

#### Pull request checklist


packages/shared/src/scanner/scanner.ts
```diff
@@ -113,9 +122,16 @@ export class Scanner {
             await this.allProgressReporter.completeRun(combinedReportParameters, combinedScanResult.baselineEvaluation);
             return this.allProgressReporter.didScanSucceed();
         } catch (error) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
katydecorah, line 125: Does this file still require this line?

packages/shared/src/input-validator.ts
```diff
@@ -34,6 +52,7 @@ export class InputValidator {
             const siteDirName = this.taskConfig.getInputName('StaticSiteDir');
             const urlName = this.taskConfig.getInputName('Url');
             const errorLines = [`A configuration error has occurred, ${urlName} or ${siteDirName} must be set`];
+            this.telemetryErrorCollector.collectError(errorLines.join(sectionSeparator()));
```
katydecorah, line 55: Starting at this line, I noticed a pattern where the file calls `this.telemetryErrorCollector.collectError` and then `tis.writeConfigurationError` several times:

```
this.telemetryErrorCollector.collectError(errorLines.join(sectionSeparator()));
return this.writeConfigurationError(errorLines);
```

What do you think about updating `writeConfigurationError` method to add `this.telemetryErrorCollector.collectError(errorLines.join(sectionSeparator()));`? This would allow us to reduce the number of times we call `collectError` and help reduce potential future human error (for example, if we add a new input to be validated, we will have to make sure we also call `collectError`).

katydecorah (commented):
This is looking really good, @ThanyaLeif! I left a few comments with some suggestions and questions. 

Also, do you think it's worth adding telemetry client mocks to existing tests in `shared/src/scanner/scanner.spec.ts`? There are a few error message related tests in that file (such as "reports error when timeout occurs and returns false") and it could give us more coverage on the new telemetry event, if you agree 🙂.

packages/shared/src/scanner/scanner.ts
ThanyaLeif: Removed!

packages/shared/src/input-validator.ts
ThanyaLeif: Oh yeah, that's a great idea!
Also modify the function a bit to accept as input a `string` instead of `string[]` and make the `errorLines.join(sectionSeparator())` outside the method

packages/shared/src/telemetry/telemetry-error-collector.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    constructor(errorSender: ErrorSender) {
+        this.errorSender = errorSender;
+    }
+
+    public collectError(errorMessage: string): void {
+        this.errorList.push(errorMessage);
+    }
+
+    public cleanErrorList(): void {
```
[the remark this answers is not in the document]
ThanyaLeif: Actually, I forgot to add this method at the end of the `InputValidator` check 😬, my intention is to avoid storing any errors from this class while running `Scanner`, since InputValidator is called inside it.

katydecorah (approved):
Nice work, Tania 💐!

PR #1535, opened 2023-02-09
chore: Add browser resolution to scan summary

katydecorah:
#### Details

This pull request proposes two changes:

1. Replace hardcoded `browserResolution` with the value from `scanMetadata`. 
1. Add the browser resolution to the scan summary.

Link to canary build:
- I pushed this branch to my staging extension, here's the scan summary: https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=42010&view=ms.vss-build-web.run-extensions-tab

##### Motivation

Last year, the scan package experienced a regression where it was scanning at 800x600. Our tests did not catch this regression as the browser resolution value was hardcoded and the codebase didn't have a snapshot test to captures the value. Adding the browser resolution to the scan summary will allow us to track this value (with e2e test) while also providing users with extra metadata about their scan.

#### Pull request checklist


packages/ado-extension/e2e/ado-extension.test.ts
```diff
@@ -9,6 +9,40 @@ describe('Sample task tests', () => {
     beforeEach(() => {
         inputs = {};
     });
+
+    it('returns expected markdown report', () => {
```
katydecorah, line 13: Note: I think an e2e test might be the best way to track the actual browser resolution value pulled from the scan package. Instead of capturing just the one sentence containing the browser resolution, this test expands to the greater markdown report so we can have a higher resolution snapshot (for example, axe-core version is another value that is pulled in from elsewhere).

brocktaylor7 (approved):
Looks great! I really like this change. 

Great work, Katy! 👏

PR #1538, opened 2023-02-10
feat(telemetry-and-inputs): add staticSiteDir telemetry

brocktaylor7:
#### Details

This adds staticSiteDir to the list of inputs telemetry that we're collecting.  

Rather than collect the entire value for `staticSiteDir`, this records a `staticSiteDirSet` boolean value, as we don't need to know the path, only if they're using the input or not.

##### Motivation

We want to track how often the static site option is being used with the end goal of removing it in the future if it is not being used much. Removing this input in the future would allow us to also remove the `dynamicSite` and `staticSite` options that tend to cause confusion for many teams onboarding. It also wouldn't prevent people from scanning static sites, they would just need to serve those sites in the pipeline on their own.

#### Pull request checklist

- [x] Ran prechecking (`yarn prechecking`)

### DIFF
--- a/packages/ado-extension/src/task-config/ado-task-config.ts
+++ b/packages/ado-extension/src/task-config/ado-task-config.ts
@@ -32,14 +32,12 @@ export class ADOTaskConfig extends TaskConfig {
         return this.memoizedReportOutDir;
     }
 
-    public getStaticSiteDir(): string {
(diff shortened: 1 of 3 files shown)


ThanyaLeif (approved):
🚢

PR #1559, opened 2023-02-24
feat: migrate from global yarn v1 to bundled yarn v3

### LINKED ISSUE #1109: Reduce onboarding complexity by replacing dependency on yarn
## Describe the bug

This issue reflects some conversation on our support thread and with @dbjorge 

`yarn` is a an undocumented dependency of the Azure DevOps extension task. This is because the task installs its dependencies at runtime using the `yarn` package manager. While Microsoft hosted pool agents include `yarn` by default, some build environments don't have `yarn` installed. This increases onboarding complexity because users need to debug the error and add `yarn` to their pipeline themselves.

Users also need `node >= 10.x`. This is because the puppeteer install script (and perhaps others) run 'node install.js'. The ADO task execution handler runs node10 but installing node6 for example, may prefix the PATH and fail - puppeteer uses async syntax which isn't supported until 10.x. 

### Constraints

We install dependencies at runtime instead of pre-building them for a few reasons:
- VSIX files have a maximum size limit of 50 MB. [The documentation](https://docs.microsoft.com/en-us/azure/devops/extend/publish/overview?view=azure-devops#check-package-size) suggests a mitigation: "Fetch things at runtime or during install time rather than providing it within the package."
- Some deps like `puppeteer` have different postinstall steps based on the OS platform; we might need to create and maintain several bundles

### Current approach

We maintain a `yarn.lock` file that represents the required runtime dependencies. We assume `yarn` exists on the build agent and [invoke `yarn install`](https://github.com/microsoft/accessibility-insights-action/blob/main/packages/ado-extension/src/install-runtime-dependencies.ts) towards the beginning of task execution.

### Possible improvements

- [x] #1123
- [ADO documentation](https://docs.microsoft.com/en-us/azure/devops/extend/publish/overview?view=azure-devops#check-package-size) also suggests "Consider using the tool installer lib to pull tool dependencies at runtime. Using the lib offers benefits where the tool is cached by version so for private agents, it won't get downloaded every build. We made it a lib so it can be used outside of tool installer tasks. But, the task won't work in disconnected scenarios (no internet), which should be in the description / docs for the task." See if leveraging the tool installer lib ([sample here](https://github.com/microsoft/azure-pipelines-tool-lib/blob/master/sample.ts)) would help us.
- Switch from `yarn` to `npm` for runtime-installation since `npm` will exist more often in `node` environments than `yarn` will. We don't use `npm` for our own local development yet because of issues like https://github.com/npm/cli/issues/4232#issuecomment-1034240038 - we need override behavior. If we take this approach, Dan suggests converting our `yarn.lock` file to `package-lock.json` using something like https://github.com/imsnif/synp.
- ...others?

### LINKED ISSUE #1551: Accessibility-insights for azure devops not working and throw exception
## Describe the bug

The accessibility task was working very well but suddnly it start report this error:
_node:internal/modules/cjs/loader:1210
  return process.dlopen(module, path.toNamespacedPath(filename));_

image

Added a new task to install NodeJS version 16 but unfortunately we got a different error:
ERROR PuppeteerCrawler: handleRequestFunction failed, reclaiming failed request back to the list or queue {"url":"[https://shopin3d-ppe.azurefd.net/en-us/store/shop-in-3d/surfaceprox.html","retryCount":1,"id":"HuyCQqyoK9MYjRm"}](https://shopin3d-ppe.azurefd.net/en-us/store/shop-in-3d/surfaceprox.html%22,%22retryCount%22:1,%22id%22:%22HuyCQqyoK9MYjRm%22%7D)
  Error: Navigation timed out after 60 seconds.
      at handleRequestTimeout (D:\a\_tasks\accessibility-insights_4811a442-2fd3-5aa8-ba1a-14cb7e24c113\3.2.0\node_modules\apify\build\crawlers\crawler_utils.js:19:11)
      at PuppeteerCrawler._handleNavigationTimeout (D:\a\_tasks\accessibility-insights_4811a442-2fd3-5aa8-ba1a-14cb7e24c113\3.2.0\node_modules\apify\build\crawlers\browser_crawler.js:418:54)
      at PuppeteerCrawler._handleNavigation (D:\a\_tasks\accessibility-insights_4811a442-2fd3-5aa8-ba1a-14cb7e24c113\3.2.0\node_modules\apify\build\crawlers\browser_crawler.js:401:18)
      at runNextTicks (node:internal/process/task_queues:61:5)
      at processTimers (node:internal/timers:499:9)
      at async PuppeteerCrawler._handleRequestFunction (D:\a\_tasks\accessibility-insights_4811a442-2fd3-5aa8-ba1a-14cb7e24c113\3.2.0\node_modules\apify\build\crawlers\browser_crawler.js:343:13)
      at async wrap (D:\a\_tasks\accessibility-insights_4811a442-2fd3-5aa8-ba1a-14cb7e24c113\3.2.0\node_modules\@apify\timeout\index.js:73:27)
image

**To Reproduce**

2. Click on '...'
3. Scroll down to '...'

- OS Name & Version: 
- Azure DevOps Extension Version & Environment: 
- Browser Version: 
- Target Page:

dbjorge:
#### Details

This PR migrates the repo to yarn v3, both for development of the extension and also for our customers. Notable impacts of this are:

* A copy of yarn 3 is now checked into the repo
* The same copy of yarn 3 is now bundled with the ADO extension and invoked directly. This means that users will no longer need to include any explicit `npm install -g yarn` step, and it insulates us from issues due to from differences in yarn versions based on customer environment.
* Because of this above, it is easier to invoke our bundled yarn using the same node binary that our script runs within. This means that users will no longer need to include any explicit `NodeTools@0` step, and it insulates us from issues due to differences in Node versions based on customer environment (I suspect this should fix #1551).
* We get to drop our `lerna` dependency in favor of the yarn workspaces plugin, which substantially reduces our overall dependency tree size
* Pre-empts the possibility of package-squatting attacks against workspace-internal packages

##### Motivation

* Removes a few steps from onboarding
* Prevents several classes of errors we've seen customers hit recently
* Enables more precise yarn resolutions
* Pre-empts package-squatting attacks

##### Context

* [Service repo's yarn v3 migration PR for reference](https://github.com/microsoft/accessibility-insights-service/pull/2210)
* [Yarn's v1 -> v2+ migration guide that I followed for the first few commits](https://yarnpkg.com/getting-started/migration)

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1109


package.json
```diff
@@ -55,7 +57,7 @@
         "jest-extended": "^3.2.4",
         "jest-file-snapshot": "^0.5.0",
         "jest-junit": "^15.0.0",
-        "lerna": "^6.5.1",
+        "js-yaml": "^4.1.0",
```
dbjorge, line 60: `js-yaml` is here because bundling a working `.yarnrc.yml` required manipulating our base one a bit at runtime. I picked this over other yaml libraries (particularly, `yaml`) by asking "among widely-used and recently-maintained options, which one do we already have the latest version of in our transitive dependency tree anyway?"

katydecorah:
@dbjorge Overall, I think this would be a big improvement to the onboarding flow!

A few small questions:

- What does the maintenance of having yarn checked into the repo look like? Should we expect to update this periodically?
- Curiosity question: I noticed that the package.json files now have added `devDependencies`, was that a result of switching from lerna to yarn 3?
- I think this pull request could remove `lerna.json` at the root, correct?

dbjorge:
> * What does the maintenance of having yarn checked into the repo look like? Should we expect to update this periodically?

Yes, we should expect to update this periodically, probably at a pace comparable to updating node versions we use.

> * Curiosity question: I noticed that the package.json files now have added `devDependencies`, was that a result of switching from lerna to yarn 3?

This is a result of yarn v2+ being more strict about importing/using dependencies that aren't explicitly listed. Particularly, yarn v2+ doesn't place entries in `packages/subpackage/node_modules/.bin` for devDependencies that aren't explicitly listed by the subpackage, which breaks most dev tools we use in practice.

> * I think this pull request could remove `lerna.json` at the root, correct?

Yep, good catch, will push an update.

dbjorge:
[Test run against 8e89098](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=42614&view=logs&j=7fa645d6-8008-59f5-c288-e237730992a6&t=9fc10df1-2a6a-5124-3d07-6f31ed1ddf6b). Ran a matrix of test cases (all combinations of hosted agent images `ubuntu-18.04`, `ubuntu-20.04`, `ubuntu-22.04`, `windows-2019`, `windows-2022` each with all combinations of "no explicit node/yarn install steps", "install node 16", "install node 16 + yarn", "install node 18"), all worked as expected.

katydecorah (approved):
This will be a big onboarding improvement! Nice work, @dbjorge!

Issue #1575, opened 2023-03-03
ADO extension fails when run in a container:  handleRequestFunction failed, { message: 'read ECONNRESET' }


katydecorah:
## Describe the bug

When run in a container, the ADO extension fails with: 

```
 handleRequestFunction failed, reclaiming failed request back to the list or queue
```

It will sometimes includes these additional error messages:
- ` ProtocolError: Protocol error (Target.setAutoAttach): Target closed.`
- `caused by: Error: "{ message: 'read ECONNRESET' }"`

**To Reproduce**

You can reproduce the error by running the extension in a container. https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#single-job

```yaml
pool:
  vmImage: 'windows-2019'

container: mcr.microsoft.com/windows/servercore:ltsc2019

steps:
- script: set

- task: NodeTool@0
  inputs:
      versionSpec: '16.x'

- script: npm install yarn@1.22.10 -g
  displayName: install yarn as a global dependency
    
- task: accessibility-insights.prod.task.accessibility-insights@3
  displayName: '[should fail] staging accessibility insights task against URL with known failures'
  inputs:
    url: 'https://www.washington.edu/accesscomputing/AU/before.html'
```

## Repro example

- https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=42790&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7384d774-f7ca-599c-ee57-ab2c05be9247
- https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=42788&view=logs&j=6b902995-b73d-5f5c-66fd-a7f66c857d2c&t=ff9c4bdf-81d4-57c3-e146-7d6c292d3dad

## Expected behavior

According to [Running Puppeteer in Docker, Troubleshooting](https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#running-puppeteer-in-docker) there are likely extra steps we should recommend to the user.

It's not clear to me if there's anything we can do on the extension side or if this can be helped with added documentation

## Context (please complete the following information)

- OS Name & Version: 
- Azure DevOps Extension Version & Environment: 
- Browser Version: 
- Target Page: 

## Are you willing to submit a PR?

Yes

dbjorge:
[...]

PR #1759, opened 2023-08-05
fix: Emit a warning when on an unsupported Windows platform

codeofdusk:
#### Details
Some Windows Server versions (Nano Server and Server Core) lack the needed dependencies to run Chrome. This PR adds a check for this case (based on the information in https://github.com/microsoft/accessibility-insights-action/issues/1575#issuecomment-1528353206), displaying a warning in the console when it occurs. Specifically, it checks for the availability of the `Server-Media-Foundation` feature (only available on server editions, not client) with non-installed state.

##### Motivation
Addresses #1575.

##### Context
This pr adds a new dependency ([wmi-client](https://www.npmjs.com/package/wmi-client)) to `package.json`. In order to use this package and pass linting, we needed to create a declaration file for its types. This is `packages/ado-extension/src/wmic.d.ts`, which declares the `"wmi-client"` `module` and its `wmic` `class`.

#### Pull request checklist

- [x] Addresses an existing issue: #1575 (partially, only for Windows)


packages/ado-extension/src/install-runtime-dependencies.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
codeofdusk, line 1: I'm unsure of how to specify type hints around calls into the `node-wmi` package. Will add type hints before taking this out of draft.

packages/ado-extension/src/install-runtime-dependencies.ts
```diff
@@ -33,5 +34,28 @@ export function installRuntimeDependencies(): void {
...
+        const wmi = new WMIC();
+
+        wmi.query(
+            `SELECT Caption,InstallState FROM Win32_OptionalFeature WHERE name="ServerMediaFoundation"`,
+            function (err: string, features: Feature[]) {
+                const INSTALLSTATE_INSTALLED = 1;
+                for (const feat of features) {
+                    if (feat['InstallState'] !== INSTALLSTATE_INSTALLED)
+                        console.log(
+                            `##[warning]${feat['Caption']} is not installed! Verify that you are using a supported image (Server Core and Nano Server are not supported).`,
```
DaveTryon, line 54: What's the correlation between this warning going off and the failure occurring? If it's 100%, I'd suggest that it would be better to error out here, as opposed to forcing the user to comb through the logs to try to understand why it failed

DaveTryon (approved):
Approved with a question/suggestion

packages/ado-extension/src/install-runtime-dependencies.ts
madalynrose: @codeofdusk I defer to you on this one!

codeofdusk:
After discussion (offline) with @DaveTryon:

> It's an improvement over what we currently have. If people still can't figure it out, we can make it stronger

Issue #1713, opened 2023-06-15
Error: "pQuerySelectorAll is not a function"


DaveTryon:
## Describe the bug

A customer has reported that version 3.4.0 of the ADO Extension is failing with the following (redacted) log:
```
Starting: Scan for accessibility issues in Canary
==============================================================================
Task         : Accessibility Insights Azure DevOps Task
Description  : Scan for accessibility issues in an Azure DevOps pipeline
Version      : 3.4.0
Author       : Accessibility Insights
Help         : For help with configuring and troubleshooting this task, see https://aka.ms/accessibility-insights-faq#azure-devops-extension
==============================================================================
Installing runtime dependencies
##[debug][Telemetry] tracking a 'ScanStart' event
##[debug][Telemetry] tracking a 'Inputs' event
##[debug][Telemetry] tracking a 'AuthUsed' event
Scanning URL [REDACTED]
##[debug]Starting accessibility scanning of URL [REDACTED]
##[debug]Chrome app executable: /usr/bin/google-chrome
##[debug] Created a data storage folder at /home/vsts/work/_temp/accessibility-insights-action-bj4suj. You can override the path by setting the APIFY_LOCAL_STORAGE_DIR environment variable
##[debug] PuppeteerCrawler:AutoscaledPool:Snapshotter: Setting max memory of this run to 1736 MB. Use the APIFY_MEMORY_MBYTES environment variable to override it.
##[debug] PuppeteerCrawler:AutoscaledPool: state {"currentConcurrency":0,"desiredConcurrency":1,"systemStatus":{"isSystemIdle":true,"memInfo":{"isOverloaded":false,"limitRatio":0.2,"actualRatio":null},"eventLoopInfo":{"isOverloaded":false,"limitRatio":0.6,"actualRatio":null},"cpuInfo":{"isOverloaded":false,"limitRatio":0.4,"actualRatio":null},"clientInfo":{"isOverloaded":false,"limitRatio":0.3,"actualRatio":null}}}
##[debug]KMSI page handled.
##[debug]Authentication succeeded.
Processing [REDACTED]
ERROR PuppeteerCrawler: handleRequestFunction failed, reclaiming failed request back to the list or queue {"url":"[REDACTED","retryCount":1,"id":"EpDaKLo3wIzVW6z"}](REDACTED%22,%22retryCount%22:1,%22id%22:%22EpDaKLo3wIzVW6z%22})
[...]

PR #1824, opened 2023-09-21
chore: Reorder validation steps in pipeline template

DaveTryon:
#### Details

The current validation steps are 

##### Motivation

Make it easier to determine the successful and failing cases. The cases are called out as succeeding or failing and are separately numbered to make it easier to add new cases to each in the future.

The `success-3` case is currently failing due to #1713, so we will need to update the template once that is fixed. The final result can be found at https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=47465&view=results

##### Context

The commits here are intentionally small, which might help simplify the review process

#### Pull request checklist

- [n/a] Addresses an existing issue: 


madalynrose (approved):
Looks good!

PR #1874, opened 2023-10-16
Chore: Make validation pipelines more sensitive and meaningful

DaveTryon:
#### Details

Our validation pipeline is not very good at surfacing customer-impacting failures. It intentionally contains "should succeed" and "should fail" test cases, but failures from the "should fail" cases could mean "we completed the analysis and found accessibility errors" or "we failed to complete the analysis". ADO returns both cases as "partially successful", and the differences are only apparent by drilling down into the logs. This is why the puppeteer failure went unnoticed for so long.

This PR improves validation usefulness by splitting the validation into 2 jobs--one that runs the tests, and one that validates the artifacts generated while running those tests. If any of the test cases fail to run, it will cause the set of artifacts to change, and the validation job will fail. If the validation job fails, then ADO will report the overall build status as "failed" instead of "partially succeeded".

The validation job intentionally checks only the _presence_ of the expected artifact files, without validating the _contents_ of the expected artifact files. This allows us to continue to use the same template across Canary, Insider, and Production, even if the artifact content varies due to changes in our code or in a package that we consume (examples: ai-scan, axe-core). Artifact content validation is already handled in the PR and CI builds, where the code and the tests can be kept tightly synchronized.

This PR also reclassifies the azure portal test from "should pass" (a classification added when we weren't 100% sure of the actual status due to the puppeteer problem) to "should fail" because it scans a page with a known accessibility error. It also puts the "should fail" tests that generate no artifacts at the end of the list of "should fail" tests that _do_ generate artifacts. This results in more intuitive naming for the "should fail" artifacts.

Validation build using this change is at https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=48232&view=results.

Review note: This PR involves adding some additional indentation to the old job. It's a lot easier if you select the "Hide whitespace" option when reviewing. It might also be useful to review it on a per-commit basis. The reclassification went in first, followed by adding the validation job.

##### Motivation

Make it easier to detect customer-impacting failures via the validation pipeline

#### Pull request checklist

- [n/a] Addresses an existing issue: Fixes

### DIFF
--- a/pipelines/ado-extension-validation-template.yaml
+++ b/pipelines/ado-extension-validation-template.yaml
@@ -41,17 +41,6 @@ steps:
           outputArtifactName: 'accessibility-reports-case-succeed-2'
           baselineFile: '$(System.DefaultWorkingDirectory)/dev/website-baselines/up-to-date-5858.baseline'
 
-    # The continueOnError for this case reflects a known issue with the task that will be fixed in a future release
-    - task: ${{ parameters.taskUnderTest }}
-      displayName: '[should succeed] case succeed-3: url, azure portal auth test'
-      inputs:
-          url: 'https://portal.azure.com'
-          serviceAccountName: $(a11yinsscan-service-acct-username)
-          serviceAccountPassword: $(a11yinsscan-service-acct-password)
-          authType: 'AAD'
-          outputArtifactName: 'accessibility-reports-case-succeed-3'
-      continueOnError: true
-
     # Please keep all "should fail" cases together as a block after all "should succeed" cases
     # Each case will need to have continueOnError: true. Each case but the first will also need
     # a condition: succeededOrFailed() so that the pipeline will run correctly.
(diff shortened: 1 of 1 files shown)


pipelines/ado-extension-validation-template.yaml
```diff
@@ -15,126 +15,182 @@ parameters:
...
+                    echo artifacts/accessibility-reports-case-fail-4/with-resolved-failures-5858.baseline >> expected-baseline-files.txt
+                    echo artifacts/accessibility-reports-case-fail-5/doesnt-exist.baseline >> expected-baseline-files.txt
+                    echo Expected .baseline files \(expected-baseline-files.txt\):
+                    cat expected-baseline-files.txt
+                    echo 
+                    echo Actual .baseline files from artifacts \(actual-baseline-files.txt\):
+                    find artifacts -name *.baseline | sort | tee actual-baseline-files.txt
+                    echo 
+                    echo Comparing actual-baseline-files.txt to expected-baseline-files.txt
+                    diff actual-baseline-files.txt expected-baseline-files.txt
```
codeofdusk, line 195: Tested this separately, looks like `diff` exits 1 when the files are different. Cool!

codeofdusk (approved):

pipelines/ado-extension-validation-template.yaml
DaveTryon: Added

PR #2056, opened 2024-04-02
feat(crawl-hash-urls): Updated ADO task validation pipeline for new parameter

v-sharmachir:
#### Details

Added the new task in validation pipeline to validate the new parameter "keepUrlFragment" functionality in pipeline

##### Motivation

[Feature 2114928](https://dev.azure.com/mseng/1ES/_workitems/edit/2114928): ADO Extension should be able to crawl hash-routing SPA apps

#### Pull request checklist

### DIFF
--- a/pipelines/ado-extension-validation-template.yaml
+++ b/pipelines/ado-extension-validation-template.yaml
@@ -26,9 +26,17 @@ jobs:
             inputs:
                 versionSpec: '16.x'
                 displayName: Use Node 16.x
-          # reused by all "url" cases
+
+          - script: yarn install --immutable
+            displayName: 'yarn install'
+    
+          # reused by all "url" cases except hash routing
           - script: npx serve "$(System.DefaultWorkingDirectory)/dev/website-root" -l 5858 &
             displayName: 'Start /dev/website-root test server at http://localhost:5858'
+
+            # reused by all "url" cases with hash routing
+          - script: yarn workspace test-website-hashroute start &
+            displayName: 'Start /packages/test-website-hashroute test server'
 
           # Please keep all "should succeed" cases together as a block before and "should fail" cases.
           # Do not specify a condition so that the pipeline will fail if any of these fail.
@@ -136,6 +144,25 @@ jobs:
             condition: succeededOrFailed()
             continueOnError: true
 
+          # Below cases are for keepUrlFragment parameter validation
+          - task: ${{ parameters.taskUnderTest }}
+            displayName: '[should crawl] case keepUrlFragment-1: crawl to hash routed page'
+            inputs:
+                url: 'http://localhost:3000/'
+                keepUrlFragment: 'true'
+                outputArtifactName: 'accessibility-reports-case-keepUrlFragment-1'
+            condition: succeededOrFailed()
+            continueOnError: true
+
+          - task: ${{ parameters.taskUnderTest }}
+            displayName: '[should not crawl] case keepUrlFragment-2: do not crawl to hash routed page'
+            inputs:
+                url: 'http://localhost:3000/'
+                keepUrlFragment: 'false'
+                outputArtifactName: 'accessibility-reports-case-keepUrlFragment-2'
+            condition: succeededOrFailed()
+            continueOnError: true
+
     - job: validate_artifacts
       dependsOn: run_tests
       displayName: 'Validate artifacts'
@@ -160,6 +187,8 @@ jobs:
                     echo artifacts/accessibility-reports-case-fail-5/index.html >> expected-index-html-files.txt
                     echo artifacts/accessibility-reports-case-fail-6/index.html >> expected-index-html-files.txt
                     echo artifacts/accessibility-reports-case-fail-7/index.html >> expected-index-html-files.txt
+                    echo artifacts/accessibility-reports-case-keepUrlFragment-1/index.html >> expected-index-html-files.txt
+                    echo artifacts/accessibility-reports-case-keepUrlFragment-2/index.html >> expected-index-html-files.txt
                     echo artifacts/accessibility-reports-case-succeed-1/index.html >> expected-index-html-files.txt
                     echo artifacts/accessibility-reports-case-succeed-2/index.html >> expected-index-html-files.txt
                     echo artifacts/accessibility-reports/index.html >> expected-index-html-files.txt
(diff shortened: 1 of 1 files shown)


pipelines/ado-extension-validation-template.yaml
```diff
@@ -192,3 +221,28 @@ jobs:
...
+                script: |
+                    cd artifacts/accessibility-reports-case-keepUrlFragment-1
+                    grep -r --include='*.html' '5</span> total URLs scanned' > accessibility-reports-case-keepUrlFragment-1.txt
+                    if [ -s "accessibility-reports-case-keepUrlFragment-1.txt" ]; then
+                    echo test-passed > expected-result.txt
+                    echo keepUrlFragment-1-test-passed
+                    fi
+
+                    cd ../accessibility-reports-case-keepUrlFragment-2
+                    grep -r --include='*.html' '5</span> total URLs scanned' > accessibility-reports-case-keepUrlFragment-2.txt
```
v-viyada, line 240: This should be 1 because when flag is set to false, it should only scan the home url.

v-viyada (changes requested):

pipelines/ado-extension-validation-template.yaml
v-sharmachir: changed the no. of URL scanned along the below if condition

pipelines/ado-extension-validation-template.yaml
```diff
@@ -192,3 +221,28 @@ jobs:
...
+            displayName: 'Validate keepUrlFragment from index.html file'
+            # Below validation is based on number of URLs scanned by crawler in index.html file
+            # with keepUrlFragment as false, only 1 Url should be scanned
+            # with keepUrlFragment as true, 5 Urls should be scanned
+            inputs:
+                script: |
+                    cd artifacts/accessibility-reports-case-keepUrlFragment-1
+                    grep -r --include='*.html' '5</span> total URLs scanned' > accessibility-reports-case-keepUrlFragment-1.txt
+                    if [ -s "accessibility-reports-case-keepUrlFragment-1.txt" ]; then
+                    echo test-passed > expected-result.txt
```
DaveTryon, line 235: Should the lines in the if block be indented?

pipelines/ado-extension-validation-template.yaml
```diff
@@ -192,3 +221,28 @@ jobs:
...
+                    echo keepUrlFragment-1-test-passed
+                    fi
+
+                    cd ../accessibility-reports-case-keepUrlFragment-2
+                    grep -r --include='*.html' '1</span> total URLs scanned' > accessibility-reports-case-keepUrlFragment-2.txt
+                    if [ -s "accessibility-reports-case-keepUrlFragment-2.txt" ]; then
+                    echo test-passed > expected-result.txt
```
DaveTryon, line 242: Should the lines in the if block be indented?

pipelines/ado-extension-validation-template.yaml
```diff
@@ -192,3 +221,28 @@ jobs:
...
+
+          - task: CmdLine@2
+            displayName: 'Validate keepUrlFragment from index.html file'
+            # Below validation is based on number of URLs scanned by crawler in index.html file
+            # with keepUrlFragment as false, only 1 Url should be scanned
+            # with keepUrlFragment as true, 5 Urls should be scanned
+            inputs:
+                script: |
+                    cd artifacts/accessibility-reports-case-keepUrlFragment-1
+                    grep -r --include='*.html' '5</span> total URLs scanned' > accessibility-reports-case-keepUrlFragment-1.txt
```
DaveTryon, line 233: This validation is hardcoded to the HTML formatting. It feels brittle, but since it should only change when the scanner is updated, maybe it's ok?

pipelines/ado-extension-validation-template.yaml
```diff
@@ -192,3 +221,28 @@ jobs:
...
+                    fi
+
+                    cd ../accessibility-reports-case-keepUrlFragment-2
+                    grep -r --include='*.html' '1</span> total URLs scanned' > accessibility-reports-case-keepUrlFragment-2.txt
+                    if [ -s "accessibility-reports-case-keepUrlFragment-2.txt" ]; then
+                    echo test-passed > expected-result.txt
+                    echo keepUrlFragment-2-test-passed
+                    fi
+
+                    cd ..
```
DaveTryon, line 246: Would a comment about diff's behavior make sense here? Specifically, it needs to fail if neither file exists

pipelines/ado-extension-validation-template.yaml
```diff
@@ -26,10 +26,18 @@ jobs:
             inputs:
                 versionSpec: '16.x'
                 displayName: Use Node 16.x
-          # reused by all "url" cases
+
+          - script: yarn install --immutable
```
DaveTryon, line 30: The yarn install was not needed before... what has changed to make it necessary now?

v-sharmachir: we added a test app with hash routing for e2e and pipeline testing. To download the dependencies and start that app yarn install is required now

pipelines/ado-extension-validation-template.yaml
```diff
@@ -26,10 +26,18 @@ jobs:
...
+            displayName: 'yarn install'
+
+          # reused by all "url" cases except hash routing
           - script: npx serve "$(System.DefaultWorkingDirectory)/dev/website-root" -l 5858 &
             displayName: 'Start /dev/website-root test server at http://localhost:5858'
 
+            # reused by all "url" cases with hash routing
```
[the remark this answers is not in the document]
v-sharmachir: fixed

pipelines/ado-extension-validation-template.yaml
v-sharmachir: fixed

pipelines/ado-extension-validation-template.yaml
v-sharmachir: fixed

pipelines/ado-extension-validation-template.yaml
v-sharmachir: added the note

pipelines/ado-extension-validation-template.yaml
v-sharmachir: I found URL count would be the consistent way to check crawling. Since in html file tag is there in between URL count and other text, I tried other approach as well and left with no other option but to keep it. Please suggest if any other alteration is required or it is good as is.
image

pipelines/ado-extension-validation-template.yaml
DaveTryon: Here's my concern: Customers won't be expected to run `yarn install`, so running it creates a mismatch between what we test and what customers use. That's far from ideal. I can think of a few options:
1. Rather than deploying some app locally, test an SPA app than already exists and is highly reliable (perhaps https://portal.azure.com). Specify the max # of pages that should be crawled, both to limit the time and to provide a guide of how many pages should be found. Firewall limitations may or may not allow this option, although we're only running in 1 pipeline and can potentially work around any problems that exist.
2. If local deployment is mandatory, maybe there's a way to deploy the app without running `yarn install`. If the project is MIT licensed, maybe we make a copy of it in our repo and host it from there. That reasonably approximates what customers do and should be fairly easy.
3. If `yarn install` turns out to be mandatory, perhaps we can minimize the potential for mismatch. We could run the non-SPA tests without `yarn install` by running the SPA tests in a separate job or by delaying the `yarn install` until _after_ the non-SPA tests have completed.

Bottom line: Let's keep the validation environment as close as possible to the customer environment

v-sharmachir: Removed the yarn install script and updated the validation based on the URL https://ms.portal.azure.com instead of using the local test app.  Max # of pages that should be crawled is set to 10

DaveTryon (approved):

PR #2082, opened 2024-05-09
chore: update publishing pipelines to use federated credentials instead of PATs and client secrets

madalynrose:
#### Details

Our release pipelines for the ADO extension utilize a Visual Studio Marketplace Service Connection with a PAT to publish to the Visual Studio Marketplace. This PR creates a new template task (`generate-vs-marketplace-token.yaml`) that uses AzureCLI to generate an access token for that service connection to use instead. This template is then used directly before the tasks that involve the VS Marketplace SC (the `QueryVersion` task in `package-vsix-file.yaml` and the `PublishAzureDevOpsExtension` task in `publish-vsix-file.yaml`). 

Additionally, this PR implements the new ESRP signing task that uses managed identity with federated credentials instead of client secrets to authenticate.

It also deletes the yaml file for the pipeline that creates a new ESRP service connection since that type of service connection is no longer used in the signing task and we won't need to regenerate the entire SC during secret rotation anymore.

##### Motivation

Security improvements

##### Context

This solution for removing PAT usage in the VS Marketplace Publishing step was based on [this workflow](https://jessehouwing.net/publish-azure-devops-extensions-using-workload-identity-oidc/), linked from [this issue](https://github.com/microsoft/azure-devops-extension-tasks/issues/506) in microsoft/azure-devops-extension-tasks.  

These changes coincide with the addition of the following new variables to the pipeline in ADO (documenting here in case they ever need to be changed):
`linuxImage`: the name of the linux image in the hosted pool to be used for this pipeline (currently set to `ubuntu-22.04-secure`)
`MARKETPLACE_RESOURCE_ID`: the resource ID of the VS Marketplace SC
`esrp-app-registration-client-id`: the client ID of the ESRP Signing app registration 
`esrp-app-registration-tenant-id`: the tenant where the ESRP Signing app registration lives

#### Pull request checklist


pipelines/generate-vs-marketplace-token.yaml
```diff
@@ -0,0 +1,15 @@
...
+steps:
+    - task: AzureCLI@2
+      displayName: Get accessToken
+      name: getAccessToken
+      inputs:
+          azureSubscription: a11y-insights-action-prod
+          scriptType: pscore
+          scriptLocation: 'inlineScript'
+          inlineScript: |
+              $accessToken = az account get-access-token --resource 499b84ac-1321-427f-aa17-267ca6975798 --query "accessToken" --output tsv
```
brocktaylor7, line 13: Is this the azure resource id? Is there any concern with having the actual guid here?

pipelines/generate-vs-marketplace-token.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+steps:
+    - task: AzureCLI@2
+      displayName: Get accessToken
+      name: getAccessToken
+      inputs:
+          azureSubscription: a11y-insights-action-prod
```
v-viyada, line 9: Is it going to be same for canary, staging and prod environments?

madalynrose: Adding answer for posterity: yes!

pipelines/generate-vs-marketplace-token.yaml
madalynrose: I've added a comment in the code to address this.

